PR #10, opened 2021-12-27
improve backend OpenProject URL validation

### LINKED ISSUE #9: hard to setup for a simple "test-environment"
connecting this app to a simple OpenProject test-environment / development environment running e.g. under `http://localhost:3000` is prevented by:
1. replacing the openproject URL to have `https` as protocol
2. checking the openproject URL to have at least one dot in the domain name regex: `'/^(https?:\/\/)?[^.]+\.[^.].*/'`
3. NC not accepting `localhost` as "remote server" - this is easily solvable by setting a system setting `./occ config:system:set allow_local_remote_servers --value 1`

would you be happy for me to do a PR changing 1. and 2. ? 

For 1. we could have a warning saying it is not "secure" using "HTTP"
For 2. I think a domain name without a `.` should be just accepted because its perfectly valid

individual-it:
followup to c4d6540aa3bf8f8fbd8748a08281207ba6fb7189
the validation has also to be fixed in the backend code 

instead of using a regex I propose to use the PHP build-in `filter_var` function and a simple regex to make sure the URI starts with http(s) as protocol

added unit tests to make sure it works correctly

fixes #9

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -1,0 +1,11 @@
+<?php
+
+use Composer\Autoload\ClassLoader;
+
+include_once __DIR__.'/vendor/autoload.php';
+
+$classLoader = new ClassLoader();
+$classLoader->addPsr4("OCA\\OpenProject\\Service\\", __DIR__ . '/lib/Service', true);
+$classLoader->register();
+
+set_include_path(get_include_path() . PATH_SEPARATOR . '/usr/share/php');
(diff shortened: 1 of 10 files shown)


eneiluj (changes requested):
Works fine, thanks!

I feel like the `validateOpenProjectURL()` static method would be better placed in `OpenProjectAPIService`. Would you mind moving it there?

individual-it:
@eneiluj done, that even made the tests a bit simpler, please review again

eneiluj (approved):

PR #11, opened 2021-12-28
accept URL for oAuth with insignificant differences

individual-it:
if the URL entered in the admin settings and in the personal settings are different only by a `/` the app does not allow to connect via oAuth. It took me a while to figure out what is wrong when I tried to setup oAuth.
This PR strips the `/` of the end of the URLs when comparing them and adds some Vue unit tests to check if the oAuth button in shown correctly


eneiluj (approved):
:+1:

PR #13, opened 2021-12-28
improve OAuth documentation

individual-it:
- rename labels be to consistent with OpenProject
  to reduce confusion use the same term as OpenProject 
  grafik
- improve OAuth documentation
  more details about how to set it up

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,25 @@ A link to the "Connected accounts" user settings section will be displayed in th
 
 ### Admin settings
 
-There also is a "Connected accounts" **admin** settings section if you want to allow your Nextcloud users to use OAuth to authenticate to a specific OpenProject instance. An admin can create an OAuth app (and get a client ID and a client secret) on OpenProject side in Administration -> Authentication -> OAuth applications.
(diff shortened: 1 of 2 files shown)


eneiluj (approved):

lindenthal:
@individual-it and @eneiluj thanks for improving the documentation. I just tried to understand how both settings are related:

1. Personal settings using the personal API token
2. Global setting usint the oAuth application

Can you please revise the documentation again and provide a documentation for less tech savy users? Maybe the following questions help:

- What is the difference between the authentication using the personal API token and the authorization using the oAuth-flow?
- Does every user in my organization has to enter their personal API token in their personal setttings?
- What permissions do I need to setup the oAuth-application in OpenProject?
- What permissions do I need to setup the oAuth-application in Nextcloud?
- ...

individual-it:
@lindenthal 
the admin can set a global setting, but that is not enough to setup oauth. Every single user has to enter also a openproject URL, IF that is the same as the admin entered in the global settings, then the user will be presented with an oauth button.
Still the user can setup a different server and use a personal token.

I will revise the docs again

PR #16, opened 2021-12-29
[Rebased] accept URL for oAuth with insignificant differences

eneiluj:
I just rebased the branch of #11 on master.
@individual-it Could you check the resolved conflicts in Makefile and package.json are ok?


individual-it (approved):
Looks good to me

PR #19, opened 2022-01-18
OP#40534 option for admins to prevent users to setup individual API-token based integration

individual-it:
https://community.openproject.org/work_packages/40534


individual-it:
closing in favor of #35

PR #21, opened 2022-01-24
show 'no connection' message in files tab OP#4068

individual-it:
In case there is no connection to OpenProject or workpackages could not be fetched, the user should see that information in the files tab and be able to navigate to the connection settings from there.

Currently this will always show "Error" as fetching workpackages is not implemented yet

### DIFF
--- a/src/components/tab/EmptyContent.vue
+++ b/src/components/tab/EmptyContent.vue
@@ -4,17 +4,51 @@
 			<img :src="flowSvg" alt="flow">
 		</div>
 		<div class="title text-center">
-			{{ t('integration_openproject', "No OpenProject links yet") }}
+			{{ emptyContentMessage }}
+		</div>
+		<br>
+		<div v-if="showConnectButton" class="connect-button">
+			<a class="button" :href="settingsUrl">
+				{{ t('integration_openproject', 'Connect to OpenProject') }}
+			</a>
 		</div>
 	</div>
 </template>
 
 <script>
+import { generateUrl } from '@nextcloud/router'
+import { translate as t } from '@nextcloud/l10n'
+
 export default {
 	name: 'EmptyContent',
+	props: {
+		state: {
+			type: String,
+			required: true,
+			default: 'ok',
+		},
+	},
+	data() {
+		return {
+			settingsUrl: generateUrl('/settings/user/connected-accounts'),
+		}
+	},
 	computed: {
 		flowSvg() {
 			return require('../../../img/flow.svg')
+		},
+		showConnectButton() {
+			return ['error', 'no-token'].includes(this.state)
+		},
+		emptyContentMessage() {
+			if (this.state === 'no-token') {
+				return t('integration_openproject', 'No OpenProject account connected')
+			} else if (this.state === 'error') {
+				return t('integration_openproject', 'Error connecting to OpenProject')
+			} else if (this.state === 'ok') {
+				return t('integration_openproject', 'No workspaces linked yet')
+			}
+			return 'invalid state'
 		},
 	},
 }
--- a/src/views/ProjectsTab.vue
+++ b/src/views/ProjectsTab.vue
@@ -22,12 +22,16 @@
 
 <template>
 	<div class="projects">
-		<EmptyContent />
+		<div :class="{ 'icon-loading': state === 'loading' }">
+			<EmptyContent v-if="state !== 'loading'" id="openproject-empty-content" :state="state" />
+		</div>
 	</div>
 </template>
 
 <script>
 import EmptyContent from '../components/tab/EmptyContent'
+import axios from '@nextcloud/axios'
+import { generateUrl } from '@nextcloud/router'
 
 export default {
 	name: 'ProjectsTab',
@@ -36,8 +40,8 @@ export default {
 	},
 	data: () => ({
 		error: '',
-		loading: false,
 		fileInfo: null,
+		state: 'loading',
 	}),
 	computed: {},
 	created() {},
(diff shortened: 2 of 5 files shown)


src/views/ProjectsTab.vue
```diff
@@ -49,13 +53,33 @@ export default {
...
 			this.loading = false
 			this.error = ''
+			this.state = 'loading'
+		},
+
+		async fetchWorkpackages(fileId) {
+			const req = {}
+			const url = generateUrl('/apps/integration_openproject/workpackages/' + fileId)
+			await axios.get(url, req).then((response) => {
+				if (!Array.isArray(response.data)) {
```
individual-it, line 71: more validation need to be added later

src/views/ProjectsTab.vue
```diff
@@ -49,13 +53,33 @@ export default {
...
 		 */
 		resetState() {
 			this.loading = false
 			this.error = ''
+			this.state = 'loading'
+		},
+
+		async fetchWorkpackages(fileId) {
+			const req = {}
+			const url = generateUrl('/apps/integration_openproject/workpackages/' + fileId)
```
individual-it, line 69: this route is not implemented yet, so it will always return `200` with the HTML content of the dashboard

tests/jest/views/ProjectsTab.spec.js
```diff
@@ -0,0 +1,71 @@
...
+	beforeEach(() => {
+		wrapper = shallowMount(ProjectsTab, { localVue })
+	})
+	describe('loading icon', () => {
+		it('shows the loading icon during "loading" state', async () => {
+			wrapper.setData({ state: 'loading' })
+			await localVue.nextTick()
+			expect(wrapper.find(loadingIndicatorSelector).exists()).toBeTruthy()
+		})
+		it('does not shows the empty content message during "loading" state', async () => {
```
phil-davis, line 22:
```suggestion
		it('does not show the empty content message during "loading" state', async () => {
```

phil-davis (approved):
LGTM - I only found one tiny grammar error in a test description!

tests/jest/views/ProjectsTab.spec.js
individual-it: fixed

individual-it:
OP#4068

src/components/tab/EmptyContent.vue
```diff
@@ -4,18 +4,49 @@
 			<img :src="flowSvg" alt="flow">
 		</div>
 		<div class="title text-center">
-			{{ t('integration_openproject', "No OpenProject links yet") }}
+			{{ emptyContentMessage }}
+		</div>
+		<br>
+		<div v-if="state === 'no-token' || state === 'error'" class="connect-button">
```
kiranparajuli589, line 10:
```suggestion
		<div v-if="['error', 'no-token].includes(state)" class="connect-button">
```
you can also use this logic from the `computed` section, just to make HTML clean
```
<div v-if="someComputedMethod" class="connect-button">
```

src/components/tab/EmptyContent.vue
```diff
@@ -4,18 +4,49 @@
...
 		flowSvg() {
 			return require('../../../img/flow.svg')
 		},
+		emptyContentMessage() {
+			if (this.state === 'no-token') {
+				return t('integration_openproject', 'No OpenProject account connected')
+			} else if (this.state === 'error') {
+				return t('integration_openproject', 'Error connecting to OpenProject')
+			} else if (this.state === 'ok') {
+				return t('integration_openproject', 'No OpenProject links yet')
```
kiranparajuli589, line 46:
```suggestion
				return t('integration_openproject', 'No workspaces linked yet')
```

src/views/ProjectsTab.vue
```diff
@@ -49,13 +53,33 @@ export default {
...
+			await axios.get(url, req).then((response) => {
```
kiranparajuli589, line 70: if you're using `then` & `catch`, await may not be necessary

src/views/ProjectsTab.vue
```diff
@@ -49,13 +53,33 @@ export default {
 		 */
 		async update(fileInfo) {
 			this.fileInfo = fileInfo
+			await this.fetchWorkpackages(this.fileInfo.id)
 		},
 		/**
 		 * Reset the current view to its default state
 		 */
 		resetState() {
 			this.loading = false
```
kiranparajuli589, line 62: this can also be removed

tests/jest/views/ProjectsTab.spec.js
```diff
@@ -0,0 +1,71 @@
...
+			wrapper.setData({ state: 'loading' })
+			await localVue.nextTick()
+			expect(wrapper.find(loadingIndicatorSelector).exists()).toBeTruthy()
+			wrapper.setData({ state })
+			await localVue.nextTick()
+			expect(wrapper.find(loadingIndicatorSelector).exists()).toBeFalsy()
+		})
+	})
+	describe('empty message', () => {
+		it.each(['no-token', 'ok', 'error'])('shows the empty message on states != loading', async (state) => {
```
kiranparajuli589, line 37:
```suggestion
		it.each(['no-token', 'ok', 'error'])('should show the empty message when state is other than loading', async (state) => {
```

src/views/ProjectsTab.vue
eneiluj: I agree with @kiranparajuli589 . It's clearer to choose between `await` in a try/catch block and `.then().catch()` without `await`.

As async/await is used all the way up to `projectTab.js::mount()`, this could just become
``` js
try {
    const response = await axios.get(url, req)
    if (!Array.isArray(response.data)) {
        ...
    }
} catch (error) {
    ...
}
```

eneiluj (commented):
LGTM.

I agree with all comments from @kiranparajuli589.

src/components/tab/EmptyContent.vue
individual-it: done

src/views/ProjectsTab.vue
individual-it: done

src/views/ProjectsTab.vue
individual-it: I think we need that line because we need to fetch the information every time an other file is selected

tests/jest/views/ProjectsTab.spec.js
individual-it: done

src/components/tab/EmptyContent.vue
individual-it: done

individual-it:
@kiranparajuli589 @eneiluj thank you for the reviews, I have addressed the comments. 
Please review again
For the easy to re-review I've created a new commit - when this is ready to be merged I will squash it.

src/views/ProjectsTab.vue
kiranparajuli589: since `loading` from `data` section is removed, now `this.loading` means nothing. it will only warn you about the variable not found.

individual-it: :facepalm: I thought you commented on `await this.fetchWorkpackages(this.fileInfo.id)`
you are totally right, I've removed the line

kiranparajuli589 (approved):
lgtm :+1:

eneiluj (approved):

PR #22, opened 2022-01-25
test OAuth requests using pact.io as mock server OP#40804

individual-it:
using the framework of pact.io to mock the OAuth requests of OpenProject and testing the `request` function

I'm open for discussion if that is the framework we want to use to mock the requests. The advantage is that, we can simulate real requests and responses without having to spin up a complete NC or OpenProject server.
The framework writes the requests and response into a JSON file and in future we can use that to verify that what we expect from OpenProject is really what they send.

here a short intro in how it works: https://docs.pact.io/getting_started/how_pact_works#consumer-testing

### DIFF
--- a/tests/lib/Service/OpenProjectAPIServiceTest.php
+++ b/tests/lib/Service/OpenProjectAPIServiceTest.php
@@ -9,9 +9,73 @@
 
 namespace OCA\OpenProject\Service;
 
+use GuzzleHttp\Client as GuzzleClient;
+use OC\Http\Client\Client;
+use OCP\ICertificateManager;
+use OCP\IConfig;
+use OCP\ILogger;
+use PhpPact\Consumer\InteractionBuilder;
+use PhpPact\Consumer\Model\ConsumerRequest;
+use PhpPact\Consumer\Model\ProviderResponse;
+use PhpPact\Standalone\MockService\MockServerEnvConfig;
 use PHPUnit\Framework\TestCase;
 
-class OpenProjectAPIServiceTest extends TestCase {
+class OpenProjectAPIServiceTest extends TestCase
+{
+	/**
+	 * @var InteractionBuilder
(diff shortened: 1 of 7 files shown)


wielinde:
Hi everybody,

I am not familiar with the PHP toolchain etc. So I cannot be of great help here. At OP we use similar stuff, for instance [_Puffing Billy_](https://github.com/oesmith/puffing-billy), which allows for mocking services on the browser level. It definitely has its legitimacy.

Needless to say, that it will not save us from:
- Manual tests
- Real integration tests
which we need to tackle and setup one by one, too.

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -39,8 +102,83 @@ public function urlsDataProvider(): array {
...
+	public function testValidateOpenProjectURL(string $url, bool $expected)	{
 		$result = OpenProjectAPIService::validateOpenProjectURL($url);
 		$this->assertSame($expected, $result);
 	}
+
+	public function testRequestUsingOAuthToken() {
+		$consumerRequest = new ConsumerRequest();
+		$consumerRequest
+			->setMethod('GET')
+			->setPath('/api/v3/work_packages')
```
kiranparajuli589, line 114: using a variable for this path will be useful since it is used multiple times below.

eneiluj (approved):
I'm fine with any choice you make for testing.

tests/lib/Service/OpenProjectAPIServiceTest.php
individual-it: done, moved the path and also the URL of the mock-server into a variable, please review again

kiranparajuli589 (commented):
looks good :+1:

kiranparajuli589 (approved):

PR #24, opened 2022-01-27
unit tests for  OpenProjectAPIService::getNotifications() OP#40804

individual-it:
basic unit tests for `OpenProjectAPIService::getNotifications()`
for that I also moved getting time into seperate function to make it easier to fake the time

### DIFF
--- a/tests/lib/Service/OpenProjectAPIServiceTest.php
+++ b/tests/lib/Service/OpenProjectAPIServiceTest.php
@@ -111,6 +111,100 @@ class OpenProjectAPIServiceTest extends TestCase
 		$this->assertSame($expected, $result);
 	}
 
+	public function testGetNotificationsRequest() {
+		$consumerRequest = new ConsumerRequest();
+		$consumerRequest
+			->setMethod('GET')
+			->setPath($this->workPackagesPath)
+			->setHeaders(["Authorization" =>  "Bearer 1234567890"])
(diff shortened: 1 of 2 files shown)


tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -111,6 +111,100 @@ public function testValidateOpenProjectURL(string $url, bool $expected) {
...
+		$service->expects($this->once())
+			->method('request')
+			->with(
+				'url','token', 'type', 'refresh', 'id', 'secret','user', 'work_packages',
+				[
+					'filters' => '[{"updatedAt":{"operator":"<>d","values":["2022-01-01T12:01:01Z","2022-01-27T08:15:48Z"]}},{"status":{"operator":"!","values":["14"]}}]',
+					'sortBy' => '[["updatedAt", "desc"]]'
+			]);
+
+		$result = $service->getNotifications('url','token', 'type', 'refresh', 'id', 'secret', 'user', '2022-01-01T12:01:01Z');
```
kiranparajuli589, line 194: assert result missing :thinking: ?

individual-it: the assertion is `$service->expects` expecting the request function to be called with the correct parameters, itself will return an error but we are not really interested in it in this test I think.
The `$result` variable is useless here, I've removed it

kiranparajuli589 (approved):
lgtm :+1

PR #30, opened 2022-01-31
l10n: Correct spelling

Valdnet:
Reported by the translator at Transifex.

Signed-off-by: Valdnet <47037905+Valdnet@users.noreply.github.com>


rakekniven (commented):
"connecting" is a verb here. For me it reads pretty well. If we change this to "connection" then it needs to be "the connection" IMHO.

Valdnet:
@rakekniven Corrected 👍.

individual-it (approved):

PR #31, opened 2022-01-31
Small style adjustments

eneiluj:
`.fill-height` is only present in `<EmptyContent>` so it might be cleaner to define its style there.

`.center-content` style is already set in `<ProjectsTab>` and this affects the child components. No need to define it in `tab.css`.

In general I would say it's best to define the style in the components and only use `tab.css` when it's the only way.
What do you think?

### DIFF
--- a/css/tab.css
+++ b/css/tab.css
@@ -9,12 +9,6 @@ body.theme--dark span.icon-open-project {
 	background: url('../img/tab-logo-white.svg') no-repeat center;
 }
 
-#tab-open-project, .fill-height {
+#tab-open-project {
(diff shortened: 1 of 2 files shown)


kiranparajuli589 (approved):
> In general I would say it's best to define the style in the components and only use `tab.css` when it's the only way. What do you think?

I totally agree with you @eneiluj. whenever there is a need for CSS styles used in multiple components, it's better to implement them inside `tab.css`.

Sorry that I pushed duplicated pieces ;) thanks for the cleanup! :+1:

PR #32, opened 2022-02-01
Reuse app svg in different usages

kiranparajuli589:
### Description
- Only use one `app.svg` (use css filter to manage light/dark theme) everywhere
- ~~Only use one `tab-logo.svg` (use css filter to manage light/dark theme)~~

### Related
OP#40085


lib/Search/OpenProjectSearchProvider.php
```diff
@@ -115,10 +115,10 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
 
 		$theme = $this->config->getUserValue($user->getUID(), 'accessibility', 'theme');
-		$thumbnailUrl = ($theme === 'dark')
-			? $this->urlGenerator->imagePath(Application::APP_ID, 'app.svg')
-			: $this->urlGenerator->imagePath(Application::APP_ID, 'app-dark.svg');
-
+//		$thumbnailUrl = ($theme === 'dark')
+//			? $this->urlGenerator->imagePath(Application::APP_ID, 'app.svg')
+//			: $this->urlGenerator->imagePath(Application::APP_ID, 'app-dark.svg');
+		$thumbnailUrl = $this->urlGenerator->imagePath(Application::APP_ID, 'app.svg');
```
kiranparajuli589, line 121: @eneiluj can we provide `CSS` class here rather providing the SVG

eneiluj: AFAIK providing a class is not possible.
But we have an endpoint to get svg files with a fill color. I'll check if it's possible to use that in there.

eneiluj: I pushed 2 commits to this branch:
* search result icon URL is now the SVG generator endpoint which accepts a stroke/fill color parameter
* use app.svg as sidebar tab icon and delete tab-logo.svg
@kiranparajuli589 What do you think?

kiranparajuli589: thanks, that's great! I was also wondering why using different SVG for tabs!

kiranparajuli589:
@eneiluj the pr is ready I think. please review

eneiluj (approved):
:+1:

PR #33, opened 2022-02-02
Set light svg to pure white

kiranparajuli589:
### Description
- fixes strange color in app list for the integration app
- color used in the SVG is made pure white
- NC uses a filter to make that SVG a bit grey for a light/dark theme, we do not have to manage that (we just have to provide a pure white SVG)
- the same SVG is used in the dashboard widget. the changes make no damage there.

### Related
OP#40085

### Screenshot
- App List (light)
    Screenshot from 2022-02-02 12-26-04
- App List (dark)
    Screenshot from 2022-02-02 12-26-32
- Dashboard Widget
    Screenshot from 2022-02-02 12-32-49


kiranparajuli589:
should be fixed with https://github.com/nextcloud/integration_openproject/pull/32

PR #35, opened 2022-02-03
OP#40534 only allow user to connect via OAuth

individual-it:
the user should not have any option to connect via api tokens to OpenProject. This PR is part of that
- the oAuth URL is constructed in PHP and handed over to vue
- oAuth button is a separate Vue component and used in files-details-tab, dashboard, personal settings
-  in personal settings page the user can only connect and disconnect from OpenProject (via oAuth)

in following PR more work needs to be done:
- delete back-end code that deals with api tokens
- show correct status of the connection in the file-details-tab (currently connect button is displayed even the connection is established)

https://community.openproject.org/projects/817/work_packages/40534/activity


src/components/OAuthConnectButton.vue
```diff
@@ -0,0 +1,51 @@
...
+					window.location.replace(requestUrl)
+				})
+				.catch((error) => {
+					showError(
+						t('integration_openproject', 'Failed to save OpenProject OAuth state')
+						+ ': ' + error.message
+					)
+				})
+				.then(() => {
+				})
```
kiranparajuli589, line 47: `finally` maybe?

src/components/PersonalSettings.vue
```diff
@@ -15,47 +12,12 @@
...
-			<button v-if="showOAuth && !connected"
-				id="openproject-oauth"
-				:disabled="loading === true"
-				:class="{ loading }"
-				@click="onOAuthClick">
-				<span class="icon icon-external" />
-				{{ t('integration_openproject', 'Connect to OpenProject') }}
-			</button>
+			<OAuthConnectButton v-if="!connected" :request-url="requestUrl" />
 			<div v-if="connected" class="openproject-grid-form">
```
kiranparajuli589, line 16:
```suggestion
			<div v-else class="openproject-grid-form">
```

src/components/OAuthConnectButton.vue
individual-it: completely deleted because nothing to be done in the second `then`

src/components/PersonalSettings.vue
individual-it: I've moved the logic around and placed the `OAuthConnectButton` at the bottom with an `v-else`

kiranparajuli589 (approved):

individual-it:
I've not adjusted also the documentation

lib/Settings/Personal.php
```diff
@@ -1,9 +1,11 @@
 <?php
 namespace OCA\OpenProject\Settings;
 
+use OCA\OpenProject\Service\OpenProjectAPIService;
```
eneiluj, line 4: Just nitpicking here :grin:
Let's keep the `OCA\OpenProject` imports grouped lower if you don't mind.

lib/Settings/Personal.php
```diff
@@ -22,45 +24,39 @@ class Personal implements ISettings {
 	 * @var string|null
 	 */
 	private $userId;
+	private IURLGenerator $url;
```
eneiluj, line 27: Php 7.3 compat

lib/Service/OpenProjectAPIService.php
```diff
@@ -491,4 +492,23 @@ public static function validateOpenProjectURL(string $openprojectUrl): bool {
...
+
+	/**
+	 * @param IConfig $config
+	 * @param IURLGenerator $urlGenerator
+	 * @return string
+	 * generates an oauth url to OpenProject containing client_id & redirect_uri as parameter
+	 * please note that the state parameter is still missing, that needs to be generated dynamically
+	 * and saved to the DB before calling the OAuth URI
+	 */
+	public static function getOpenProjectOauthURL(IConfig $config, IURLGenerator $urlGenerator): string {
```
eneiluj, line 504: `IURLGenerator` could be injected in the constructor of `OpenProjectAPIService` instead of passing it as a parameter to `getOpenProjectOauthURL`. This would avoid having to inject `IURLGenerator` in every class calling this method.

eneiluj (changes requested):
Very nice.
A few minor inline comments.

lib/Service/OpenProjectAPIService.php
individual-it: yes, but then every class that calls the method would have to initiate the complete `OpenProjectAPIService` and by that need to get injected even more
e..g from `OpenProjectWidget` we would need the logger, usermanager, avatarmanager etc.
Or maybe we want to place that function somewhere else? A helper class?

lib/Dashboard/OpenProjectWidget.php
```diff
@@ -32,17 +35,21 @@
...
-	/** @var IL10N */
-	private $l10n;
-	/**
-	 * @var IURLGenerator
-	 */
-	private $url;
+	private IL10N $l10n;
+	private IURLGenerator $url;
+	private IInitialState $initialStateService;
+	private IConfig $config;
```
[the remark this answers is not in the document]
individual-it: thx for the hint done in 9316b6b100dc90e4971f37583fa381dddd4159c2

lib/Listener/LoadSidebarScript.php
```diff
@@ -28,16 +28,37 @@
...
 use OCP\EventDispatcher\IEventListener;
+use OCP\IConfig;
+use OCP\IURLGenerator;
 use OCP\Util;
 
 class LoadSidebarScript implements IEventListener {
+
+	private IURLGenerator $url;
+	private IInitialState $initialStateService;
+	private IConfig $config;
```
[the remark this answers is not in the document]
individual-it: done in 9316b6b100dc90e4971f37583fa381dddd4159c2

lib/Settings/Personal.php
individual-it: done in 9316b6b100dc90e4971f37583fa381dddd4159c2

lib/Settings/Personal.php
individual-it: those things are good to fix and keep in order, done in dd9f2cc0f7578b7e721e60df97725701ea68b41c
hope that code style analysis will help in future

individual-it:
@eneiluj could you re-review and give your opinion on my comment here https://github.com/nextcloud/integration_openproject/pull/35#discussion_r799203671 ?

lib/Service/OpenProjectAPIService.php
```diff
@@ -560,4 +561,23 @@ public function getOpenProjectWorkPackageType(
...
+		$clientID = $config->getAppValue(Application::APP_ID, 'client_id');
+		$oauthUrl = $config->getAppValue(Application::APP_ID, 'oauth_instance_url');
+		$redirectUri = $urlGenerator->linkToRouteAbsolute(Application::APP_ID . '.config.oauthRedirect');
+		return $oauthUrl .
+			'/oauth/authorize' .
+			'?client_id=' . $clientID .
+			'&redirect_uri=' . $redirectUri .
```
eneiluj, line 580:
```suggestion
			'&redirect_uri=' . urlencode($redirectUri) .
```

lib/Service/OpenProjectAPIService.php
eneiluj: Yes I agree. As long as `getOpenProjectOauthURL` is a static method, let's keep it that way.
If you feel like it belongs in a new class/file, feel free to move it :grin:.

eneiluj (dismissed):
:+1:
One last small change and we're good to go.

lib/Service/OpenProjectAPIService.php
individual-it: good point, I've adjusted that and also squashed all commits

individual-it:
I found an other issue that I've missed before. The url to openproject was still taken from the user settings

PR #38, opened 2022-02-04
[#40963] PHP codestyle in CI

individual-it:
- run PHP-CS checks in CI
- fix PHP code style according to @nextcloud coding standards https://github.com/nextcloud/coding-standard/


kiranparajuli589 (approved):
looks good :+1:

PR #42, opened 2022-02-11
[OP#40868] development environment with docker compose

kiranparajuli589:
### Description
Setup developer environment for the integration app with `docker compose`

### Related
OP#40868

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,14 +69,16 @@ Requirements:
 
 ### Setup
 ```shell
-# the app needs to be cloned inside the "custom_apps" dir anywhere in the host
-mkdir /dev/custom_apps
+# the app needs to be cloned inside the "custom_apps" dir
+mkdir $HOME/development/custom_apps -p
+cd $HOME/development/custom_apps
 git clone https://github.com/nextcloud/integration_openproject.git
 # installation & building
 npm ci
 npm run build
-# provide ownership of "custom_apps" to the user "www-data"
-sudo chown www-data custom_apps -R
+# provide group ownership of "custom_apps" to the user "www-data"
+sudo chgrp www-data $HOME/development/custom_apps -R
+sudo chmod g+w $HOME/development/custom_apps -R
 ```
 
 ### Environments
@@ -98,14 +100,14 @@ After this, you should be able to access nextcloud server at [http://localhost](
 > **Note:** These steps will only be necessary for the first setup.
 
 #### Create admin
+1. browse to [http://localhost](http://localhost)
+2. create admin user
+3. get an installed NC server
+
 For the database, **PostgreSQL** is used with the following credentials:
 - **Database:** `nextcloud`
 - **User:** `nextcloud`
 - **Password:** `nextcloud`
-
-1. browse to [http://localhost](http://localhost)
-2. create admin user
-3. get an installed NC server
 
 #### Enable integration app 
 You can browse as admin to the apps center and enable it using the webUI. Or, you can just use the terminal as:
@@ -124,6 +126,6 @@ docker exec --user www-data integration_openproject_nc php occ config:system:set
 Now you can watch for the app code changes using the following command and start developing.
 
 ```shell
-cd /dev/custom_apps/integration_openproject
+cd $HOME/development/custom_apps/integration_openproject
 npm run watch
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -57,3 +57,70 @@ The account configuration happens in the "Connected accounts" user settings sect
...
+- OpenProject server instance running in the host machine
+- OpenProject Integration app
+
+### Environments
+- **APP_DIR**
+  - description: location where the `integration_openproject` repository is cloned 
+  - default: `./../integration_openproject`
+
+### Start compose
+**Note:** If your host has running apache or nginx, you need to stop it before starting the `docker-compose`.
```
individual-it, line 76: can't we use an alternative port inside of docker, so it would not affect any server on the host?

README.md
```diff
@@ -57,3 +57,70 @@ The account configuration happens in the "Connected accounts" user settings sect
...
+You can browse as admin to the apps center and enable it using the webUI. Or, you can just use the terminal as:
+
+```shell
+docker exec --user www-data integration_openproject_nc php occ a:e integration_openproject
+```
+
+If you want to allow local remote servers: 
+
+```shell
+docker exec --user www-data integration_openproject_nc php occ config:system:set allow_local_remote_servers --value 1
```
individual-it, line 116: can that be done automatically?

docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+      - POSTGRES_DB=nextcloud
+      - POSTGRES_USER=nextcloud
+      - POSTGRES_PASSWORD=nextcloud
+
+  redis:
+    image: redis:alpine
+    container_name: integration_openproject_redis
+    restart: always
+    ports:
+      - "6379:6379"
```
individual-it, line 22: can't we run NC without redis?

individual-it (changes requested):

README.md
```diff
@@ -67,16 +67,28 @@ Requirements:
 - OpenProject server instance running in the host machine
 - OpenProject Integration app
 
+### Setup
+```shell
+# the app needs to be cloned inside the "custom_apps" dir anywhere in the host
+mkdir /dev/custom_apps
```
individual-it, line 73: this is confusing because there is an actual `/dev` (for devices) folder on linux
```suggestion
mkdir ~/development/custom_apps
```

README.md
```diff
@@ -67,16 +67,28 @@ Requirements:
...
+git clone https://github.com/nextcloud/integration_openproject.git
```
individual-it, line 74:
```suggestion
cd ~/development/custom_apps
git clone https://github.com/nextcloud/integration_openproject.git
```

README.md
```diff
@@ -67,16 +67,28 @@ Requirements:
...
+# installation & building
+npm ci
+npm run build
+# provide ownership of "custom_apps" to the user "www-data"
+sudo chown www-data custom_apps -R
```
individual-it, line 79: if you do that, then the developer will not have write access to the app anymore
I would suggest to give it to the group `www-data` and give write access to the group
```suggestion
# provide group ownership of "custom_apps" to the user "www-data"
sudo chgrp www-data ~/development/custom_apps -R
sudo chmod g+w ~/development/custom_apps -r
```

README.md
```diff
@@ -110,17 +114,16 @@ You can browse as admin to the apps center and enable it using the webUI. Or, yo
...
 ```shell
 docker exec --user www-data integration_openproject_nc php occ config:system:set allow_local_remote_servers --value 1
 ```
 
 ### Start Developing
-Now you can watch for the app code changes using:
+Now you can watch for the app code changes using the following command and start developing.
 
 ```shell
+cd /dev/custom_apps/integration_openproject
```
individual-it, line 127:
```suggestion
cd ~/development/custom_apps/integration_openproject
```

individual-it (changes requested):

README.md
kiranparajuli589: when using the network mode as `host`, we cannot do port mapping

README.md
kiranparajuli589: NC server should be in installed state, so we've to do it manually

docker-compose.yml
kiranparajuli589: I've tried to run NC without Redis, works fine.
removed this service

individual-it (approved):

PR #43, opened 2022-02-11
[OP-40679] Feature search workpackages

SwikritiT:
This PR

- adds a search bar to the sidebar tab
- Implements the API for searching work packages
- Implements the UI part to list the work packages
- adds a feature file to describe how the feature should work
- adds unit tests for new feature

The implementation looks like this
final

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -66,14 +66,14 @@ class OpenProjectAPIService {
 	 * Service to make requests to OpenProject v3 (JSON) API
 	 */
 	public function __construct(
-		string               $appName,
-		IUserManager         $userManager,
-		IAvatarManager       $avatarManager,
-		LoggerInterface      $logger,
-		IL10N                $l10n,
-		IConfig              $config,
-		INotificationManager $notificationManager,
-		IClientService       $clientService) {
+								string $appName,
+								IUserManager $userManager,
+								IAvatarManager $avatarManager,
+								LoggerInterface $logger,
+								IL10N $l10n,
+								IConfig $config,
+								INotificationManager $notificationManager,
+								IClientService $clientService) {
 		$this->appName = $appName;
 		$this->userManager = $userManager;
 		$this->avatarManager = $avatarManager;
@@ -115,7 +115,7 @@ class OpenProjectAPIService {
 				// get the openproject user ID
 				$myOPUserId = $this->config->getUserValue($userId, Application::APP_ID, 'user_id');
 				if ($myOPUserId !== '') {
-					$myOPUserId = (int)$myOPUserId;
+					$myOPUserId = (int) $myOPUserId;
 					$notifications = $this->getNotifications(
 						$openprojectUrl, $accessToken, $tokenType, $refreshToken, $clientID, $clientSecret, $userId, $lastNotificationCheck
 					);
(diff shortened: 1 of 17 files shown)


lib/Service/OpenProjectAPIService.php
```diff
@@ -489,4 +489,70 @@ public static function validateOpenProjectURL(string $openprojectUrl): bool {
...
+		string $authType,
+		string $refreshToken,
+		string $clientID,
+		string $clientSecret,
+		string $userId,
+		string $statusId): array {
+		$result = $this->request(
+			$url, $accessToken, $authType, $refreshToken, $clientID, $clientSecret, $userId, 'statuses/' . $statusId);
+		if (isset($result['error'])) {
+			return $result;
```
individual-it, line 518: why do we need that, it will not do anything else than the last return in line 522

src/components/tab/SearchInput.vue
```diff
@@ -0,0 +1,308 @@
...
+							{{ workPackage.subject }}
+						</div>
+					</div>
+				</div>
+			</div>
+			<div class="create-new-wp">
+				{{ t('integration_openproject', '+ New work package in OpenProject') }}
+			</div>
+		</div>
+		<div v-if="state === 'no-token' || state === 'error' || state === 'loading' || state === 'empty'"
```
individual-it, line 52: would that also work?
```suggestion
		<div v-if="state !== 'ok'"
```

src/components/tab/SearchInput.vue
```diff
@@ -0,0 +1,308 @@
...
+				this.state = 'ok'
+			}
+		},
+		checkStatusCode(statusCode) {
+			if (statusCode === 401) {
+				this.state = 'no-token'
+			} else if (statusCode === 400) {
+				this.state = 'error'
+			} else {
+				this.state = 'error'
```
individual-it, line 128: so a `200` status code will also set the state to `error`?

src/components/tab/SearchInput.vue
```diff
@@ -0,0 +1,308 @@
...
+			}
+		},
+		replaceHrefToGetId(href, href2 = null) {
```
individual-it, line 131: I don't understand what we are trying to achieve here, can you write a comment?

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -0,0 +1,176 @@
...
+	})
+
+	describe('work packages', () => {
+		const searchListSelector = '.search-list'
+		const inputSelector = '#workpackages-search'
+		const statusSelector = '.filter-project-type-status__status'
+		const typeSelector = '.filter-project-type-status__type'
+		const assigneeSelector = '.filter-assignee'
+
+		it('should not be displayed if the length of words in searchbar is less than three', async () => {
```
individual-it, line 54:
```suggestion
		it('should not be displayed if the length of words in searchbar is less or equal than three', async () => {
```

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -0,0 +1,176 @@
...
+
+		it('should be displayed if the length of words in searchbar is more than three', async () => {
+			const wrapper = shallowMount(SearchInput, {
+				mocks: {
+					t: (msg) => msg,
+					generateUrl() {
+						return '/'
+					},
+				},
+			})
```
individual-it, line 80: can be extracted to a new function and reused

individual-it (changes requested):

src/components/tab/SearchInput.vue
```diff
@@ -0,0 +1,308 @@
...
+			if (response.status === 200) {
+				this.state = 'ok'
+				return response.data.color
+			}
+			this.checkStatusCode(response.status)
+			return ''
+		},
+		async getUserAvatar(userId, userName) {
+			const url = generateUrl('/apps/integration_openproject/avatar?')
+				+ encodeURIComponent('userId') + '=' + userId + '&' + encodeURIComponent('userName') + '=' + userName
```
individual-it, line 202: generateUrl accepts parameters
I think something like this should work:
```suggestion
			const url = generateUrl('/apps/integration_openproject/avatar', {'userId': userId, 'userName': userName )
```
but I haven't tried it

individual-it (approved):
LGTM, more improvements can be made in further PRs

PR #45, opened 2022-02-16
set background image size to fit into the element background

kiranparajuli589:
### Description
with NC server version: 23, the sidebar tab icon was completely grey.
this pr fixes the issue

### Related
OP#41083

### DIFF
--- a/css/tab.css
+++ b/css/tab.css
@@ -4,6 +4,7 @@ span.icon-openproject {
 	padding: 0 !important;
 	min-width: 44px !important;
 	min-height: 16px !important;
+	background-size: 16px 16px;
 }
 
 #tab-open-project {


individual-it:
any way we could have an automated test for that?

kiranparajuli589:
> any way we could have an automated test for that?

We can introduce visual testing.

kiranparajuli589:
failed to connect to github

https://github.com/nextcloud/integration_openproject/runs/5212034599?check_suite_focus=true

individual-it:
> > any way we could have an automated test for that?
> 
> We can introduce visual testing, but we've to set up a whole e2e environment. might be overkill just to test an icon.

that might be overkill for now

individual-it (approved):

PR #46, opened 2022-02-16
OP#40534 delete anything to do with API tokens

individual-it:
delete all code that has to do with api key based auth


kiranparajuli589 (approved):
lgtm :+1:

PR #47, opened 2022-02-16
[#40968] fix getOpenProjectAvatar error handling

individual-it:
the codepath in the fist part of the if statement could not be reached because `getOpenProjectAvatar` never returns an array with `error` as key
so I've deleted that code and covered the codepath with some basic unit tests


kiranparajuli589 (approved):
lgtm :+1:

PR #48, opened 2022-02-17
construct filters from array

individual-it:
this will make it simpler to extend the filter, e.g. when we want also to filter by `file_link_origin_id` to list all workpackages associated to the file


SwikritiT (approved):
LGTM

PR #53, opened 2022-02-21
[CI] separate coverage workflow from CI workflow

kiranparajuli589:
### Description
We have to perform checks on whether to perform coverage or not for different matrices and trigger when coverage is placed with `CI` workflow

With this pr, a separate workflow is registered just for the `coverage` part

Signed-off-by: Parajuli Kiran <kiranparajuli589@gmail.com>


kiranparajuli589:
hmm...coverage steps need coverage files from the unit test runs, so cannot be separated from the actual test run workflow.

PR #55, opened 2022-02-21
[OP#41124] Display loading icon while fetching workpackages

SwikritiT:
part of OP#41124: This PR adds a loading icon when fetching work packages

It looks something like this:
loading

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -10,6 +10,9 @@
 			track-by="multiselectKey"
 			:internal-search="false"
 			open-direction="below"
+			:loading="state === 'loading'"
+			:preselect-first="true"
+			:preserve-search="true"
 			@search-change="makeSearchRequest">
 			<template #option="{option}">
 				<div class="searchList">
@@ -87,8 +90,6 @@ export default {
 				return t('integration_openproject', 'No OpenProject account connected')
 			} else if (this.state === 'error') {
 				return t('integration_openproject', 'Error connecting to OpenProject')
-			} else if (this.state === 'loading') {
-				return t('integration_openproject', 'Wait while we fetch work packages')
 			}
 			return ''
 		},
@@ -113,11 +114,11 @@ export default {
 			}
 			const url = generateUrl('/apps/integration_openproject/work-packages')
 			if (this.search.length > 3) {
+				this.state = 'loading'
 				const req = {}
 				req.params = {
 					searchQuery: this.search,
 				}
-				this.state = 'loading'
 				const response = await axios.get(url, req)
 				if (response.status === 200) {
 					this.state = 'ok'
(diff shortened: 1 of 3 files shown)


src/components/tab/SearchInput.vue
```diff
@@ -52,6 +55,9 @@
 			<template #noOptions>
 				{{ t('integration_openproject', 'Start typing to search') }}
 			</template>
+			<template #noResult>
+				{{ noResultMessages }}
+			</template>
```
SwikritiT, line 60: @eneiluj The messages in this slot are not being displayed in the UI. The default one `No result` is displayed but if I define a custom message it's not displayed. Am I doing something wrong or that slot can't be used?

kiranparajuli589: the `#noResult` slot is actually from `vue-multiselect` library and is already implemented inside the `Multiselect.vue` so I think, we cannot further use it as a slot here.

src/components/tab/SearchInput.vue
```diff
@@ -177,10 +191,11 @@ export default {
 		async getWorkPackageColorAttributes(path, id) {
```
individual-it, line 191: this whole function seems not to be covered well with tests, maybe this PR is a good option to add more

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -145,5 +146,57 @@ describe('SearchInput.vue tests', () => {
...
 			expect(assignee.exists()).toBeFalsy()
 		})
+
+		it('should display a loading button when the work package is being fetched', async () => {
+			jest.spyOn(SearchInput.methods, 'makeSearchRequest').mockImplementation()
+			const wrapper = mountSearchInput()
+			let loadingIcon = wrapper.find(loadingIconSelector)
+			expect(loadingIcon.exists()).toBeFalsy()
+			const textInput = wrapper.find(inputSelector)
+			await textInput.setValue('organ')
```
individual-it, line 156: should not be needed if you use the `setData`

src/components/tab/SearchInput.vue
```diff
@@ -113,14 +117,15 @@ export default {
 			}
 			const url = generateUrl('/apps/integration_openproject/work_packages')
 			if (this.search.length > 3) {
+				this.loading = true
```
individual-it, line 120: why do we need this new variable and not keep on using `this.state = 'loading'`

individual-it (changes requested):

src/components/tab/SearchInput.vue
SwikritiT: The `: loading` prop of the multi-select uses a boolean value so that I defined the new variable loading that is set true or false

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -145,5 +146,54 @@ describe('SearchInput.vue tests', () => {
...
+			{
+				status: 404,
+				state: 'error',
+			},
+		])('sets states according to status', async (cases) => {
+			axios.get.mockImplementation(() =>
+				Promise.resolve({
+					data: {
+						color: 'red',
+					},
```
kiranparajuli589, line 189: `data` is also redundant here imo

SwikritiT: it gives me the error `  TypeError: Cannot read property 'color' of undefined `, if I don't define data

individual-it: can be done in an other PR, but I would suggest to improve `getWorkPackageColorAttributes`
it checks `if (response.status === 200) {` and the assumes that the response has a color attribute, but what if not?
Should be something like `if (response.status === 200 && response.data.color !== undefined) {`

individual-it (approved):

PR #56, opened 2022-02-22
[OP#41128] use PKCE with oAuth

individual-it:
use PKCE to connect  via OAuth as described in 
https://www.oauth.com/oauth2-servers/pkce/authorization-request/
and
https://www.oauth.com/oauth2-servers/pkce/authorization-code-exchange/

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -140,7 +140,7 @@ class ConfigController extends Controller {
 
 		$validCodeVerifier = false;
 		if (is_string($codeVerifier)) {
-			$validCodeVerifier = (\Safe\preg_match('/^[A-Za-z0-9\-._~]{128}$/', $codeVerifier) === 1);
+			$validCodeVerifier = (\Safe\preg_match('/^[A-Za-z0-9\-._~]{43,128}$/', $codeVerifier) === 1);
 		}
 
 		$validClientSecret = false;
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -22,6 +22,7 @@ use OCP\AppFramework\Controller;
 
 use OCA\OpenProject\Service\OpenProjectAPIService;
 use OCA\OpenProject\AppInfo\Application;
+use Psr\Log\LoggerInterface;
 
 class ConfigController extends Controller {
 
@@ -41,6 +42,12 @@ class ConfigController extends Controller {
 	 * @var OpenProjectAPIService
 	 */
 	private $openprojectAPIService;
+
+	/**
+	 * @var LoggerInterface
+	 */
+	private $logger;
+
 	/**
 	 * @var string|null
 	 */
@@ -52,12 +59,14 @@ class ConfigController extends Controller {
 								IURLGenerator $urlGenerator,
 								IL10N $l,
 								OpenProjectAPIService $openprojectAPIService,
+								LoggerInterface $logger,
 								?string $userId) {
 		parent::__construct($appName, $request);
 		$this->config = $config;
 		$this->urlGenerator = $urlGenerator;
 		$this->l = $l;
 		$this->openprojectAPIService = $openprojectAPIService;
+		$this->logger = $logger;
 		$this->userId = $userId;
 	}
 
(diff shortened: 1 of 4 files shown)


individual-it:
- [ ] ToDo: bring back the check for the secret + add tests for that see https://community.openproject.org/projects/nextcloud-integration/work_packages/41128/activity#activity-3

individual-it:
CC @Kharonus would be cool if you also could have a look on it

kiranparajuli589 (commented):
lgtm :+1:

lib/Controller/ConfigController.php
```diff
@@ -122,11 +131,24 @@ public function oauthRedirect(string $code = '', string $state = ''): RedirectRe
...
+			$this->userId, Application::APP_ID, 'code_verifier', false
+		);
 
 		// anyway, reset state
 		$this->config->deleteUserValue($this->userId, Application::APP_ID, 'oauth_state');
 
-		if ($clientID && $clientSecret && $configState !== '' && $configState === $state) {
+		$validCodeVerifier = false;
+		if (is_string($codeVerifier)) {
+			$validCodeVerifier = (\Safe\preg_match('/^[A-Za-z0-9\-._~]{128}$/', $codeVerifier) === 1);
```
Kharonus, line 143: regex should use `{43,128}`, as the verifier must be within 43 and 128 characters. I wouldn't enforce max length.

Kharonus, line 143: Where does the `$codeVerifier` comes from?

src/components/OAuthConnectButton.vue
```diff
@@ -22,14 +22,37 @@ export default {
...
+				text += possible.charAt(Math.floor(Math.random() * possible.length))
+			}
+			return text
+		},
+		async digest(text) {
+			return await crypto.subtle.digest('SHA-256', new TextEncoder().encode(text))
+		},
+		async generateCodeChallenge(codeVerifier) {
+			const digest = await this.digest(codeVerifier)
+			return btoa(String.fromCharCode(...new Uint8Array(digest)))
```
Kharonus, line 38: Is this the php way to make a base64 url encode? ^^ I'm a bit surprised by the replacement actions.

lib/Controller/ConfigController.php
Kharonus: nevermind, found it and that it is generated from caller context.

src/components/OAuthConnectButton.vue
```diff
@@ -22,14 +22,37 @@ export default {
...
+				.replace(/=/g, '')
+				.replace(/\+/g, '-')
+				.replace(/\//g, '_')
+		},
+		async onOAuthClick() {
+			const oauthState = this.generateRandomString(10)
```
Kharonus, line 44: what is the need of having this `oauthState` string here?

Kharonus (changes requested):
Just one small change request, plus 2 questions. :D

src/components/OAuthConnectButton.vue
individual-it: we need to save the state somewhere to be able to check it when requesting a token in https://github.com/nextcloud/integration_openproject/blob/a3b8ebf73e81d2425daefa9cd85a0146d310a933/lib/Controller/ConfigController.php#L121
So we generate it here and save it to the DB (in  https://github.com/nextcloud/integration_openproject/pull/56/files#diff-82450ababde1ab6945c0633d13fb1de0d6eacdd2a0e400890046c5d5a60ad34fR58) before redirecting to OpenProject 

generating the state in PHP would not work because the use could have multiple tabs open and then potentially the state send from the UI to OpenProject would not be the same as the remembered one

Kharonus: seems legitimate! 👍🏽

src/components/OAuthConnectButton.vue
individual-it: This is JS, and the replace is seem to be needed see https://github.com/tonyxu-io/pkce-generator/blob/master/index.html#L28 or https://www.oauth.com/oauth2-servers/pkce/authorization-request/

lib/Controller/ConfigController.php
individual-it: we have created a 128char string in https://github.com/nextcloud/integration_openproject/pull/56/files#diff-82450ababde1ab6945c0633d13fb1de0d6eacdd2a0e400890046c5d5a60ad34fR45
so why would be expect here anything else than 128 char?

src/components/OAuthConnectButton.vue
Kharonus: Thanks for the context ;) 👍🏽

lib/Controller/ConfigController.php
Kharonus: Because code can change in time, and there are two different locations here, that will probably not everytime change together.

This line is a check, that validates a PKCE code verifier. It could be even a separate method, or provided by a service. It should not care about how the input is generated, but stick to the RFC specification IMHO.

Kharonus (approved):
LGTM! 👍🏽

kiranparajuli589 (approved):

PR #60, opened 2022-02-24
Add composer allow plugins

kiranparajuli589:
> I updated my local system to the latest `composer` 2.2.6. When I do `composer update` it now asks me if I trust the composer plugins that we use. After answering "yes" it adds that "trust" to `composer.json`

> https://getcomposer.org/doc/06-config.md#allow-plugins
"As of Composer 2.2.0, the allow-plugins option adds a layer of security allowing you to restrict which Composer plugins are able to execute code during a Composer run."

originally posted by @phil-davis in https://github.com/owncloud/QA/issues/723

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,5 +12,10 @@
 		"cs:fix": "php-cs-fixer fix",
 		"cs:check": "php-cs-fixer fix --dry-run --diff",
 		"phpstan": "phpstan analyse"
+	},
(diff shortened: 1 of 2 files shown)


individual-it (commented):
should we update `composer.lock` in the same PR?

individual-it (approved):

PR #62, opened 2022-02-24
generate redirect_uri and don't take it from user config

individual-it:
getting it from user config is obsolete since #35 best just generate it here


kiranparajuli589 (approved):

individual-it:
Note this was missed in the original PR because I had an `redirect_uri`still in my DB so all manual tests passed.

PR #63, opened 2022-02-25
code maintenance: search input vue

kiranparajuli589:
### Description
- try-catch around Axios calls
- component state saved in constants
- improve response checks
- fix loading indicator
- updated test cases

Ref: https://v1.test-utils.vuejs.org/guides/#common-tips

### Related
https://community.openproject.org/projects/nextcloud-integration/work_packages/41295/activity
OP#41295

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -144,7 +144,7 @@ export default {
 				response = e.response
 			}
 			this.checkForErrorCode(response.status)
-			if (this.isStateOk) await this.processWorkPackages(response.data)
+			if (response.status === 200) await this.processWorkPackages(response.data)
 			if (this.isStateLoading) this.state = STATE_OK
 		},
 		async processWorkPackages(workPackages) {
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -1,7 +1,6 @@
 <template>
-	<div id="search-bar">
+	<div id="searchBar">
 		<Multiselect
-			v-model="search"
 			class="searchInput"
 			:placeholder="placeholder"
 			:options="searchResults"
@@ -10,7 +9,7 @@
 			track-by="multiselectKey"
 			:internal-search="false"
 			open-direction="below"
-			:loading="state === 'loading'"
+			:loading="isStateLoading"
 			:preselect-first="true"
 			:preserve-search="true"
 			@search-change="makeSearchRequest">
@@ -53,7 +52,7 @@
 				</div>
 			</template>
 			<template #noOptions>
-				{{ t('integration_openproject', 'Start typing to search') }}
+				{{ translate('Start typing to search') }}
 			</template>
 		</Multiselect>
 		<div v-if="state !== 'ok'"
@@ -70,6 +69,12 @@ import { translate as t } from '@nextcloud/l10n'
 import Avatar from '@nextcloud/vue/dist/Components/Avatar'
 import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
 
+const STATE_OK = 'ok'
+const STATE_ERROR = 'error'
+const STATE_NO_TOKEN = 'no-token'
+const STATE_LOADING = 'loading'
+const SEARCH_CHAR_LIMIT = 3
+
 export default {
 	name: 'SearchInput',
 	components: {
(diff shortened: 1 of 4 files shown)


src/components/tab/SearchInput.vue
```diff
@@ -177,20 +186,21 @@ export default {
...
-		async getUserAvatar(userId, userName) {
-			const url = generateUrl('/apps/integration_openproject/avatar?')
-				+ encodeURIComponent('userId') + '=' + userId + '&' + encodeURIComponent('userName') + '=' + userName
-			return url
+		getUserAvatar(userId, userName) {
+			return generateUrl('/apps/integration_openproject/avatar?')
+				+ encodeURIComponent('userId')
+				+ '=' + userId
+				+ '&' + encodeURIComponent('userName')
+				+ '=' + userName
```
individual-it, line 203: used only once can be moved to `processWorkPackages`

individual-it (changes requested):
if no tests failed in such a refactoring, we don't have enough tests, so this is an opportunity to add some

src/components/tab/SearchInput.vue
```diff
@@ -70,78 +69,56 @@ import { translate as t } from '@nextcloud/l10n'
...
-		search(value, oldValue) {
-			// if the value in the search input field changes we need to reset the searchResults
-			if (oldValue !== null) {
-				if (value.length < oldValue.length && value.length <= 3) {
-					this.searchResults = []
-					this.state = 'ok'
-				}
-			}
-		},
-	},
```
kiranparajuli589, line 107: no more required will be handled by the search event handler `makeSearchRequest`.

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -161,39 +152,170 @@ describe('SearchInput.vue tests', () => {
...
+			await inputField.setValue('org')
+
+			expect(wrapper.vm.searchResults).toMatchObject([])
+		})
+		it.each([
+			{ status: 200, expectedCallCount: 1, rejected: false },
+			{ status: 400, expectedCallCount: 0, rejected: true },
+			{ status: 401, expectedCallCount: 0, rejected: true },
+			{ status: 500, expectedCallCount: 0, rejected: true },
+		])('should call method "processWorkPackages" if only the search request succeeds',
```
individual-it, line 199:
```suggestion
		])('should call method "processWorkPackages" only if the search request succeeds',
```

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -161,39 +152,170 @@ describe('SearchInput.vue tests', () => {
...
+				await wrapper.vm.$nextTick()
+				expect(processSpy).toHaveBeenCalledTimes(expectedCallCount)
+			})
+
+		it.each([
+			{ search: 'o', expectedCallCount: 0 },
+			{ search: 'or', expectedCallCount: 0 },
+			{ search: 'org', expectedCallCount: 0 },
+			{ search: 'orga', expectedCallCount: 1 },
+		])('should send search request only if the search text is greater than search threshold', async ({ search, expectedCallCount }) => {
```
individual-it, line 227: can this be combined with 
`it('should be displayed if the length of words in searchbar is more than three', async () => {` ?

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -161,39 +152,170 @@ describe('SearchInput.vue tests', () => {
...
+			const axiosSpy = jest.spyOn(axios, 'get')
+			wrapper = mountSearchInput()
+			await wrapper.vm.makeSearchRequest(search)
+			expect(axiosSpy).toHaveBeenCalledTimes(expectedCallCount)
+		})
+		it.each([200, 400, 401, 500])('should never left the state as loading', async (status) => {
```
individual-it, line 233: duplication of `'sets states according to response status'` ?

individual-it (changes requested):

src/components/tab/SearchInput.vue
```diff
@@ -150,47 +127,64 @@ export default {
...
+			let response
+			try {
+				response = await axios.get(url, req)
+			} catch (e) {
+				response = e.response
+			}
+			this.checkStatusCode(response.status)
+			if (this.isStateOk) {
+				await this.processWorkPackages(response.data)
+			}
```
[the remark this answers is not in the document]
individual-it: https://user-images.githubusercontent.com/2425577/155944570-bb71ccfd-f794-47e1-af24-8162b58bd4bf.mp4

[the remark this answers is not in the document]
kiranparajuli589: this is due to the state update in the color attribute getter. I've removed the state update in the color attribute getter and just return an empty string if the request fails

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -86,28 +103,7 @@ describe('SearchInput.vue tests', () => {
 			expect(searchList).toMatchSnapshot()
 		})
 
-		it('should not be displayed if the length of words in searchbar decreases from more than 3 to less', async () => {
```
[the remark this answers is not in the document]
kiranparajuli589: this is a feature of `makeSearchRequest` fn, so moved to the respective `describe`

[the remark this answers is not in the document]
kiranparajuli589: this snapshot should check if all work packages options are displayed correctly.

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -56,144 +61,236 @@ describe('SearchInput.vue tests', () => {
...
-			await wrapper.setData({
-				searchResults: [],
+			it('should display correct options list of search results', async () => {
+				await wrapper.setData({
+					searchResults: workPackagesSearchResponse,
+				})
+				const searchList = wrapper.find(searchListSelector)
+
+				expect(searchList.exists()).toBeTruthy()
+				expect(searchList).toMatchSnapshot()
```
kiranparajuli589, line 91: this snapshot should check if all work packages options are displayed correctly.

tests/jest/components/tab/SearchInput.spec.js
kiranparajuli589: the test you mentioned is moved here and updated slightly.

src/components/tab/SearchInput.vue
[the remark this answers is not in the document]
kiranparajuli589: this is due to the state update after every Axios request. introduced a separate loading state to handle the mentioned case.

tests/jest/components/tab/SearchInput.spec.js
kiranparajuli589: the test you mention specifically checks the `checkStatusCode` function, but here we check for the state update after search request.

i've merged it with the test above `should call method "processWorkPackages" only if the search request succeeds`

src/components/tab/SearchInput.vue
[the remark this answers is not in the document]
individual-it: I don't think that is not needed. We have now 2 variables keeping the "same" information.
`checkStatusCode` does not need to set the state to OK, in case of `200` it can leave it as it is (`LOADING`) because that is the case, even the first request finished, the work is not done yet.
`if (this.isStateOk) {` can be changed to ether check for status `200` or allow to run `processWorkPackages` also if the state is `loading`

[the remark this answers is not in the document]
individual-it: the state `OK` should only be reached after all the work is done and not issue appeared

kiranparajuli589:
@individual-it I've fixed the loading case as discussed above. Also, I've updated the test cases. Please review again.

individual-it (approved):

PR #67, opened 2022-03-03
Add new OCS endpoint to get file info from file ID

eneiluj:
Here is one possible way to get file info from a file ID:

```
curl -H "Accept: application/json" -H "OCS-APIRequest: true" -u USER:PASSWD \
    http://my.nc.org/ocs/v1.php/apps/integration_openproject/fileinfo/FILE_ID
```

It will get the file's information from the logged in user's perspective. If the file is found multiple times (because of multiple shared accesses), one of those will be used to produce the response.

This returns something like:
``` json
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 100,
      "message": "OK",
      "totalitems": "",
      "itemsperpage": ""
    },
    "data": {
      "id": 123456,
      "name": "file.csv",
      "mtime": 1620415259,
      "ctime": 0,
      "mimetype": "text/csv",
      "path": "/Documents/file.csv",
      "size": 76521,
      "owner_name": "Jane Doe",
      "owner_id": "jane"
    }
  }
}

```

We can also make this endpoint public and use a secondary auth mechanism like temporary tokens like @individual-it suggested.


eneiluj:
New endpoint:

```
curl -H "Accept: application/json" -H "OCS-APIRequest: true" -u USER:PASSWD \
    http://my.nc.org/ocs/v1.php/apps/integration_openproject/filesinfo \
    -X POST -d '{"fileIds":[7759,6736]}' -H "Content-Type:application/json"
```

returns

``` json
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 100,
      "message": "OK",
      "totalitems": "",
      "itemsperpage": ""
    },
    "data": {
      "7759": {
        "id": 7759,
        "name": "name.md",
        "mtime": 1643102454,
        "ctime": 0,
        "mimetype": "text/markdown",
        "path": "/name.md",
        "size": 181,
        "owner_name": "Jane Does",
        "owner_id": "jane"
      },
      "6736": {
        "id": 6736,
        "name": "welcome.txt",
        "mtime": 1620415259,
        "ctime": 0,
        "mimetype": "text/plain",
        "path": "/welcome.txt",
        "size": 164,
        "owner_name": "Jane Does",
        "owner_id": "jane"
      }
    }
  }
}

```

individual-it (commented):
Looks generally good to me
I can add some tests and make CI happy if we want to integrate this

eneiluj:
> I can add some tests and make CI happy if we want to integrate this

@individual-it Yes that would be nice!

@wielinde The "multi-file" endpoint now includes the missing files. In the example of the previous comment, you can see the file with the ID 9999999 does not exist, the result then includes `"9999999: null,`.

individual-it:
ToDo:
- [ ]  tests for `getFilesInfo`
- [ ] check for `$fileIds === null` in `getFilesInfo` (that happens if an invalid JSON is send) e.g. `url -H "Accept: application/json" -H "OCS-APIRequest: true" -u  admin:admin http://localhost/nextcloud-server//ocs/v2.php/apps/integration_openproject/filesinfo -X POST -d '{"fileIds":[7759,6736,9999999' -H "Content-Type:application/json"`

individual-it:
@eneiluj I think I'm done here, can you please review if you are happy with my changes to your code. I hope I've covered all possible cases with tests

individual-it:
@wielinde what should happen if a file is deleted, and goes to the trashbin? Should this API still return the file-info?

eneiluj:
@individual-it I can't review my own PR but I'm more than fine with your changes :+1:

tests/lib/Controller/FilesControllerTest.php
```diff
@@ -0,0 +1,386 @@
...
+	/**
+	 * @return array<mixed>
+	 */
+	public function getFileInfoDataProvider() {
+		return [
+			// getById returns only one result
+			[
+				[
+					$this->getNodeMock(
+				'files/logo.png', 'logo.png', 'image/png'
```
kiranparajuli589, line 22: indent needs fixing here.

kiranparajuli589 (commented):
just found one indent issue.

tests/lib/Controller/FilesControllerTest.php
individual-it: indent fixed

individual-it (dismissed):
a file that was deleted and currently is in the trash is not returned by this API, only after restoring the file it shows up again in the result.
So if OpenProject would use this API to know if a file still exists or not, they would delete all worpkackage links of the file if its in the trashbin. After restoring the file from the trashbin there would be no way to restore the links to the workpackages

As a fix we need to check if the file is in the trashbin and if so return the files data, with an additional item saying its a deleted file, so that OpenProject would be able to react on that fact, e.g. when linking to the file from the UI

individual-it:
possible POC solution in [2c5e8cb](https://github.com/nextcloud/integration_openproject/pull/67/commits/2c5e8cb124fcdea088ad2381b33b10288ea1faaf)

individual-it:
here now how trashed results looks like:

single file:
```
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 200,
      "message": "OK"
    },
    "data": {
      "id": 3,
      "name": "welcome.txt.d1648724302",
      "mtime": 1641812049,
      "ctime": 0,
      "mimetype": "text/plain",
      "path": "/welcome.txt.d1648724302",
      "size": 164,
      "owner_name": "admin",
      "owner_id": "admin",
      "trashed": true
    }
  }
}
```

multiple files, one trashed, one existing one not existing
```
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 100,
      "message": "OK",
      "totalitems": "",
      "itemsperpage": ""
    },
    "data": {
      "3": {
        "id": 3,
        "name": "welcome.txt.d1648724302",
        "mtime": 1641812049,
        "ctime": 0,
        "mimetype": "text/plain",
        "path": "/welcome.txt.d1648724302",
        "size": 164,
        "owner_name": "admin",
        "owner_id": "admin",
        "trashed": true
      },
      "74": {
        "id": 74,
        "name": "data.zip",
        "mtime": 1625056633,
        "ctime": 0,
        "mimetype": "application/zip",
        "path": "/data.zip",
        "size": 4705,
        "owner_name": "admin",
        "owner_id": "admin",
        "trashed": false
      },
      "9999": null
    }
  }
}
```

@wielinde @Kharonus is this response OK for you?
It tells you if an item is trashed or not, but the path & name for trashed items might be strange for you. NC attached the time when the item was deleted to the items name. Also the path is internally something like `files_trashbin/files/...`. Do you want to get the whole path exposed, because you would not find `/welcome.txt.d1648724302` through the normal webDAV request

Kharonus:
@individual-it 
topic 1: pathes
i think the path like it is now is fine, as the path prefix should not be part of the path value. it is not for the files available for dav requests and should not for the trashbin files.

topic 2: deletion
When a file moved to the trashbin, the file links get not deleted. This is fine because this way we can move it out of the trashbin and restore the view within OP and nextcloud, with all the attached file links. This also means, that we would want to send a deletion request for the file links, if the files is removed from trashbin and is deleted irreversibly.

topic 3: visibility
you wrote, that having a `null` indicates a not available file. What about files, the user is not allowed to see? Are they still returned here with the meta information?

individual-it:
@Kharonus 

> topic 3: visibility
you wrote, that having a null indicates a not available file. What about files, the user is not allowed to see? Are they still returned here with the meta information?

no, only files that the user is allowed to read would return here, the API runs in the context of the authenticated user. So files that are not readable by the user would also be `null`

Kharonus:
But then we cannot distinguish between those cases. If we could distinguish them, we could identify garbage data on OP client side, and handle it. If both are null, we cannot react on it. What's your opinion on that?

individual-it:
its possible to return the fileInfo regardless of the user see 78b1e84e4d9b3e412a83b6855ef494066ca3593c
Not sure if that is the correct way of doing it @eneiluj  might advise
The issue here is now that we do leak file information to any user that guesses a fileId and those are serial

Kharonus:
Yes, this is exactly the issue. I think, it needs further clarification of how we define the behaviour of that endpoint. Not blocking for this PR though, it can be patched at a later time.

wielinde:
@individual-it @Kharonus A user should not be able to get the "meta" infos of a file that she has no access to. We should not leak it. However, telling that a file exsists or not, or if it is trashed is no leaking in my opinion.

eneiluj:
> However, telling that a file exsists or not, or if it is trashed is no leaking in my opinion.

IMO, technically it's a leak, not a very sensitive one but still. Users are not supposed to know anything about files they don't have access to.

> But then we cannot distinguish between those cases. If we could distinguish them, we could identify garbage data on OP client side, and handle it. If both are null, we cannot react on it. What's your opinion on that?

I'm sure we can find another way to let the OP client clean things up :grin:.

individual-it:
ToDo as of meeting 08.04.2022
- [ ] if the file is not accessible, return something else than `null` but no details
- [ ] the file name should always be the one the owner sees
- [ ] API tests

individual-it:
Here the newest version of the responses:
## single file 
`curl -H "Accept: application/json" -H "OCS-APIRequest: true" -u admin:admin http://localhost/nextcloud-server/ocs/v2.php/apps/integration_openproject/fileinfo/<file-id> -v`
note that here we use `v2.php` and not `v1.php` anymore, the only difference is that the HTTP statuscode are set correctly

### file exists and is readable:
- HTTP status 200
- payload:
```
 {
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 200,
      "message": "OK"
    },
    "data": {
      "id": 2,
      "name": "files",
      "mtime": 1648725686,
      "ctime": 0,
      "mimetype": "httpd/unix-directory",
      "path": "/",
      "size": 10133080,
      "owner_name": "admin",
      "owner_id": "admin",
      "trashed": false
    }
  }
}
```
### file exists and is readable but trashed:
same as above but with `trashed: true`

### file exists but is not readable
- HTTP status 403
- payload:
```
{
  "ocs": {
    "meta": {
      "status": "failure",
      "statuscode": 403,
      "message": null
    },
    "data": []
  }
}
```

### file does not exist

- HTTP status 404
- payload:
```
{
  "ocs": {
    "meta": {
      "status": "failure",
      "statuscode": 404,
      "message": null
    },
    "data": []
  }
}
```

## multifile 

`curl -H "Accept: application/json" -H "OCS-APIRequest: true" -u admin:admin     http://localhost/nextcloud-server/ocs/v1.php/apps/integration_openproject/filesinfo -X POST -d '{"fileIds":[2,3,179,999999]}' -H "Content-Type:application/json" | jq`

fileid:
- 2 exists and is readable
- 3 exists, is readable and trashed
- 179 exists but is not readable
- 999999 does not exists

result:
``` 
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 100,
      "message": "OK",
      "totalitems": "",
      "itemsperpage": ""
    },
    "data": {
      "2": {
        "id": 2,
        "name": "files",
        "mtime": 1648725686,
        "ctime": 0,
        "mimetype": "httpd/unix-directory",
        "path": "/",
        "size": 10133080,
        "owner_name": "admin",
        "owner_id": "admin",
        "trashed": false
      },
      "3": {
        "id": 3,
        "name": "welcome.txt.d1648724302",
        "mtime": 1641812049,
        "ctime": 0,
        "mimetype": "text/plain",
        "path": "/welcome.txt.d1648724302",
        "size": 164,
        "owner_name": "admin",
        "owner_id": "admin",
        "trashed": true
      },
      "179": true,
      "999999": false
    }
  }
}
```

opinions?

Kharonus:
@individual-it 
looks quite useable! Thanks for the effort. Should the `ocs.meta.status` be `200` instead of `100` on success?

I only have a small remark. I'd love to avoid booleans for state, but replace it with enums, or small objects. So, that the response is self explanatory. S.th. like

```
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 200,
      "message": "OK",
      "totalitems": "",
      "itemsperpage": ""
    },
    "data": {
      "179": "forbidden",
      "999999": null
    }
  }
}
```

or

```
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 200,
      "message": "OK",
      "totalitems": "",
      "itemsperpage": ""
    },
    "data": {
      "179": {
        "status": "failure",
        "statuscode": 403,
      },
      "999999": {
        "status": "failure",
        "statuscode": 404,
      },
    }
  }
}
```

The second maybe is more consistent even to the single file fetch. But those are not meant to be precise examples, but mere a representation of what I mean by "self explanatory responses".

individual-it:
we cannot change `ocs.meta.status` that is part of the NC API.

I like your second idea, basically having a status code for every file and those codes would be the same as HTTP status coded

individual-it:
@Kharonus what do you think of these responses:

## single file 
`curl -H "Accept: application/json" -H "OCS-APIRequest: true" -u admin:admin http://localhost/nextcloud-server/ocs/v2.php/apps/integration_openproject/fileinfo/<file-id> -v`
note that here we use `v2.php` and not `v1.php` anymore, the only difference is that the HTTP statuscode are set correctly

### file exists and is readable:
- HTTP status 200
- payload:
```
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 200,
      "message": "OK"
    },
    "data": {
      "status": "OK",
      "statuscode": 200,
      "id": 2,
      "name": "files",
      "mtime": 1648725686,
      "ctime": 0,
      "mimetype": "httpd/unix-directory",
      "path": "/",
      "size": 10133080,
      "owner_name": "admin",
      "owner_id": "admin",
      "trashed": false
    }
  }
}
```
### file exists and is readable but trashed:
same as above but with `trashed: true`

### file exists but is not readable
- HTTP status 403
- payload:
```
{
  "ocs": {
    "meta": {
      "status": "failure",
      "statuscode": 403,
      "message": null
    },
    "data": {
      "status": "Forbidden",
      "statuscode": 403
    }
  }
}
```

### file does not exist

- HTTP status 404
- payload:
```
{
  "ocs": {
    "meta": {
      "status": "failure",
      "statuscode": 404,
      "message": null
    },
    "data": {
      "status": "Not Found",
      "statuscode": 404
    }
  }
}
```

## multifile 

`curl -H "Accept: application/json" -H "OCS-APIRequest: true" -u admin:admin     http://localhost/nextcloud-server/ocs/v1.php/apps/integration_openproject/filesinfo -X POST -d '{"fileIds":[2,3,179,999999]}' -H "Content-Type:application/json" | jq`

fileid:
- 2 exists and is readable
- 3 exists, is readable and trashed
- 179 exists but is not readable
- 999999 does not exists

result:
``` 
{
  "ocs": {
    "meta": {
      "status": "ok",
      "statuscode": 100,
      "message": "OK",
      "totalitems": "",
      "itemsperpage": ""
    },
    "data": {
      "2": {
        "status": "OK",
        "statuscode": 200,
        "id": 2,
        "name": "files",
        "mtime": 1648725686,
        "ctime": 0,
        "mimetype": "httpd/unix-directory",
        "path": "/",
        "size": 10133080,
        "owner_name": "admin",
        "owner_id": "admin",
        "trashed": false
      },
      "3": {
        "status": "OK",
        "statuscode": 200,
        "id": 3,
        "name": "welcome.txt.d1648724302",
        "mtime": 1641812049,
        "ctime": 0,
        "mimetype": "text/plain",
        "path": "/welcome.txt.d1648724302",
        "size": 164,
        "owner_name": "admin",
        "owner_id": "admin",
        "trashed": true
      },
      "179": {
        "status": "Forbidden",
        "statuscode": 403
      },
      "999999": {
        "status": "Not Found",
        "statuscode": 404
      }
    }
  }
}
```

Kharonus:
@individual-it I like them. I do not mind the redundant information in the `data` and `ocs` section of the single view response.

individual-it:
I've adjusted the API, so that the path and file name that are shown are always those of the owner.
e.g.
- Brian has a file `/my-data/image.jpg`
- Brian shares `/my-data/image.jpg` to Alice
- Alice moves `image.png` to `/project-data/a-good-shot.jpg`
- When Alice requests information about the file using this API the response would be:
   ```
   {
      "status": "OK",
      "statuscode": 200,
      "id": 326,
      "name": "image.jpg",
      "mtime": 1094992774,
      "ctime": 0,
      "mimetype": "image/jpeg",
      "path": "/my-data/image.jpg",
      "size": 60936,
      "owner_name": "Brian",
      "owner_id": "brian",
      "trashed": false
    }
```
 
we would again leak some data by that, but as I understood this was agreed on

wielinde:
Hey @individual-it @Kharonus, I didn't follow the full thread here. However, do we really need the path? Where would we use it? I am asking as the path indeed leaks information that possibly should stay private, i.e. the name of some parent folder.

individual-it:
@wielinde @Kharonus yes if you want to make a WebDav request in the name of the user the path for that user would be needed

Should we display the name as the name of the owner and the path as the path of the user, or get rid of the path all together?

Kharonus:
I'd say, get rid of the path until we do need it. Then we are back to only deliver canonical information about the file, and nothing user related. Opinions?

individual-it:
@Kharonus @wielinde I think this is ready for final review. The path is gone and I've added some API tests to make sure it works with the real NC server correctly.
You can see the expected responses in tests/acceptance/features/api/getFilesinfoByFileIDsAPI.feature and tests/acceptance/features/api/getFileinfoByFileIDAPI.feature

Kharonus:
Great, LGTM. We can work with the repsonses from the acceptance tests!

individual-it (approved):

PR #64, opened 2022-02-28
[OP#41102] UI of adding link to workpackage

individual-it:
Part of: 
OP#41269
OP#41102

This PR redesigns the WorkPackage display in UI 
The new design looks like this:

Search Relation :
search -relation

Existing relation:
existing-relations

PR #68, opened 2022-03-10
[OP#41102] link workpackage backend

individual-it:
ability to link files to packages, backend for #64 
manually tested with https://github.com/opf/openproject/tree/feature/file-links

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -14,6 +14,11 @@ namespace OCA\OpenProject\Service;
 use DateTime;
 use DateTimeZone;
 use Exception;
+use OCA\OpenProject\Exception\OpenprojectErrorException;
+use OCA\OpenProject\Exception\OpenprojectResponseException;
+use OCP\Files\File;
+use OCP\Files\IRootFolder;
+use OCP\Files\NotFoundException;
 use OCP\IL10N;
 use OCP\IURLGenerator;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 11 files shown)


tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -83,10 +85,10 @@ describe('SearchInput.vue tests', () => {
 				const inputField = wrapper.find(inputSelector)
 				await inputField.setValue(search)
 				expect(axiosSpy).toHaveBeenCalledTimes(expectedCallCount)
+				axiosSpy.mockRestore()
```
individual-it, line 88: restoreAllMocks in afterEach seems not to be enough
Without this some tests would pass when running all tests of the file but fail if ran single

individual-it:
CC @Kharonus @wielinde I cannot add you as reviewers

lib/Service/OpenProjectAPIService.php
```diff
@@ -581,4 +593,94 @@ public static function getOpenProjectOauthURL(IConfig $config, IURLGenerator $ur
...
+	 * @param int $fileId
+	 * @return \OCP\Files\Node|null
+	 * @throws NotPermittedException
+	 * @throws \OC\User\NoUserException
+	 */
+	public function getFile($userId, $fileId) {
+		$userFolder = $this->storage->getUserFolder($userId);
+
+		$file = $userFolder->getById($fileId);
+		return $file[0];
```
kiranparajuli589, line 608: should we add some checks for array and index here?

individual-it: but true, the array might be empty

individual-it: I've added more checks in 3cc11ea8f409a580ec68dd2bddfe0273ae059686 and also changed the function so that it can accept also folders (using `\OCP\Files\Node` and not `\OCP\Files\File`)

kiranparajuli589 (approved):

PR #73, opened 2022-03-14
Do not show the connect button if admin config is not ok

kiranparajuli589:
## Description
For dashboard widget, sidebar tab, and personal setting:
- check if the admin config is ok (if all values are set and OAuth instance URL is valid)
- show connect button if ok
- show message `Failed to establish the connection. Please contact your administrator.` if not ok

### Related
https://community.openproject.org/projects/nextcloud-integration/work_packages/41211

### Screenshots
Screenshot from 2022-03-16 11-19-34

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -685,14 +685,21 @@ class OpenProjectAPIService {
 	}
 
 	/**
-	 * checks for every admin config variables if they are set
+	 * checks if every admin config variables are set
+	 * checks if the oauth instance url is valid
 	 *
 	 * @param IConfig $config
 	 * @return bool
 	 */
 	public static function isAdminConfigOk(IConfig $config):bool {
-		return !!($config->getAppValue(Application::APP_ID, 'client_id'))
-			&& !!($config->getAppValue(Application::APP_ID, 'client_secret'))
-			&& !!($config->getAppValue(Application::APP_ID, 'oauth_instance_url'));
+		$clientId = $config->getAppValue(Application::APP_ID, 'client_id');
+		$clientSecret = $config->getAppValue(Application::APP_ID, 'client_secret');
+		$oauthInstanceUrl = $config->getAppValue(Application::APP_ID, 'oauth_instance_url');
+		$checkIfConfigIsSet = !!($clientId) && !!($clientSecret) && !!($oauthInstanceUrl);
+		if (!$checkIfConfigIsSet) {
+			return false;
+		} else {
+			return self::validateOpenProjectURL($oauthInstanceUrl);
+		}
 	}
 }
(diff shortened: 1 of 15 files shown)


lib/Service/OpenProjectAPIService.php
```diff
@@ -683,4 +683,16 @@ public function linkWorkPackageToFile(
...
+	/**
+	 * checks for every admin config variables if they are set
+	 *
+	 * @param IConfig $config
+	 * @return bool
+	 */
+	public static function isAdminConfigOk(IConfig $config):bool {
+		return !!($config->getAppValue(Application::APP_ID, 'client_id'))
+			&& !!($config->getAppValue(Application::APP_ID, 'client_secret'))
+			&& !!($config->getAppValue(Application::APP_ID, 'oauth_instance_url'));
```
individual-it, line 696: should we also validate the URL here with `OpenProjectAPIService::validateOpenProjectURL`? or only when saving it?

src/components/tab/EmptyContent.vue
```diff
@@ -7,7 +7,8 @@
 			<div class="empty-content--title">
 				{{ emptyContentMessage }}
 			</div>
-			<div v-if="showConnectButton" class="empty-content--connect-button">
+			<br>
+			<div v-if="showConnect && showConnectButton" class="connect-button">
```
individual-it, line 11: I think this is strange naming `showConnect` and `showConnectButton`
what about doing it similar to https://github.com/nextcloud/integration_openproject/pull/73/files#diff-7f773d424a0ea444e1702d381d14bfdd2abcdbb0b199669aa28978cba4d46faeR104-R107 ?

individual-it (changes requested):
I think we also need an extra message for the user if the setting is not complete.
e.g. "OpenProject app has not been configured correctly, please contact the NextCloud administrator"

lib/Service/OpenProjectAPIService.php
kiranparajuli589: OAuth instance URL should also be valid for a successful connection. Added a check for URL too.

src/components/tab/EmptyContent.vue
kiranparajuli589: fixed

src/components/OAuthConnectButton.vue
```diff
@@ -20,7 +24,18 @@ export default {
...
 		},
+		adminConfigStatus: {
+			type: Boolean,
+			default: false,
+		},
+	},
+
+	computed: {
+		adminConfigNotOkMessage() {
+			return t('integration_openproject', 'Failed to establish connection. Please contact your administrator.')
```
individual-it, line 35:
```suggestion
			return t('integration_openproject', 'Settings of the OpenProject integration app are not complete. Please contact your NextCloud administrator.')
```

individual-it (changes requested):
just a little suggestion to change the text

individual-it (approved):

PR #75, opened 2022-03-16
refactor `getOpenProjectOauthURL` so that it throws an exception if any needed value is not set

kiranparajuli589:
### Description
- throw exception from `getOpenProjectOauthURL` if request url is not valid
- show admin settings invalid message if request URL is not valid in dashboard widget, sidebar, and personal settings components
- show/hide search input in the sidebar tab according to the request URL
- error icon when request URL is invalid for dashboard widget and empty content in the sidebar

### Screenshots
Screenshot from 2022-03-18 10-42-53

### Related
OP#41211 > 3rd todo item
https://community.openproject.org/projects/nextcloud-integration/work_packages/41211

### DIFF
--- a/src/components/PersonalSettings.vue
+++ b/src/components/PersonalSettings.vue
@@ -40,7 +40,7 @@
 					@input="onNotificationChange">
 				<label for="notification-openproject">{{ t('integration_openproject', 'Enable notifications for activity in my work packages') }}</label>
 			</div>
-			<OAuthConnectButton v-else :request-url="requestUrl" :admin-config-status="adminConfigStatus" />
+			<OAuthConnectButton v-else :request-url="requestUrl" />
 		</div>
 	</div>
 </template>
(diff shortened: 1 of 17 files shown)


src/components/PersonalSettings.vue
```diff
@@ -64,20 +64,19 @@ export default {
 		SettingsTitle, OAuthConnectButton,
 	},
 
-	props: [],
-
 	data() {
 		return {
 			loading: false,
 			state: loadState('integration_openproject', 'user-config'),
-			requestUrl: loadState('integration_openproject', 'user-config').request_url,
```
individual-it, line 73: don't we need this line for `this.requestUrl`?

individual-it (changes requested):

src/components/PersonalSettings.vue
kiranparajuli589: the request URL is a field in the `state`, so it is now returned from the computed section.

individual-it (approved):

PR #76, opened 2022-03-16
[OP#41430] periodically check openproject oauth token

individual-it:
check periodically if the oauth token is still valid (by querying `/users/me`) and if the endpoint returns an anonymous user try to refresh the token. If the refresh also failed delete all tokens and by that force the user to login again

https://user-images.githubusercontent.com/2425577/158577046-da262b31-6370-40c1-a289-c3d4ea74c563.mp4

.


individual-it:
not needed, because actually its a bug in openProject, see https://community.openproject.org/work_packages/41480

PR #78, opened 2022-03-17
[OP#41471] filter out already linked files from search

individual-it:
fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/41471

When searching for workpackages to link, don't show those that are already linked

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -55,11 +55,14 @@ export default {
 			type: Object,
 			required: true,
 		},
+		linkedWorkPackages: {
+			type: Array,
+			required: true,
+		},
 	},
 	data: () => ({
 		state: STATE_OK,
 		searchResults: [],
-		selectedId: [],
 	}),
 	computed: {
 		isStateOk() {
@@ -83,7 +86,6 @@ export default {
 	watch: {
 		fileInfo(oldFile, newFile) {
 			if (oldFile.id !== newFile.id) {
-				this.selectedId = []
 				this.resetState()
 				// FIXME: https://github.com/shentao/vue-multiselect/issues/633
 				if (this.$refs.workPackageMultiSelect?.$refs?.VueMultiselect?.search) {
@@ -123,9 +125,6 @@ export default {
 			try {
 				await axios.post(url, params, config)
 				this.$emit('saved', selectedOption)
-				this.selectedId.push({
-					id: selectedOption.id,
-				})
 			} catch (e) {
 				showError(
 					this.translate('Failed to link file to work-package')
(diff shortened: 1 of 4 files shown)


src/components/tab/SearchInput.vue
```diff
@@ -123,9 +121,6 @@ export default {
 			try {
 				await axios.post(url, params, config)
 				this.$emit('saved', selectedOption)
-				this.selectedId.push({
-					id: selectedOption.id,
-				})
```
individual-it, line 128: not needed because the data is saved directly in the parent and also compared with it when searchin

src/components/tab/SearchInput.vue
```diff
@@ -157,9 +152,9 @@ export default {
 			for (let workPackage of workPackages) {
 				try {
 					workPackage = await workpackageHelper.getAdditionalMetaData(workPackage)
-					const selectedIdFound = this.selectedId.some(el => el.id === workPackage.id)
-					const workpackageIdFound = this.searchResults.some(el => el.id === workPackage.id)
-					if (!workpackageIdFound && !selectedIdFound) {
+					const alreadyLinked = this.$parent.workpackages.some(el => el.id === workPackage.id)
+					const alreadyInSeachResults = this.searchResults.some(el => el.id === workPackage.id)
```
kiranparajuli589, line 156:
```suggestion
					const alreadyInSearchResults = this.searchResults.some(el => el.id === workPackage.id)
```

src/components/tab/SearchInput.vue
```diff
@@ -157,9 +152,9 @@ export default {
 			for (let workPackage of workPackages) {
 				try {
 					workPackage = await workpackageHelper.getAdditionalMetaData(workPackage)
-					const selectedIdFound = this.selectedId.some(el => el.id === workPackage.id)
-					const workpackageIdFound = this.searchResults.some(el => el.id === workPackage.id)
-					if (!workpackageIdFound && !selectedIdFound) {
+					const alreadyLinked = this.$parent.workpackages.some(el => el.id === workPackage.id)
```
kiranparajuli589, line 155: I would prefer using props here. although we achieve the same result, using $parent is a little vague, we don't know, whether the parent has that value or not. Also, we are able to change the value of the parent component, i.e. workpackages from the child component here, which is not "vue" way of data flow

https://stackoverflow.com/questions/53549064/vue-reasons-to-use-props-instead-of-referencing-parent-data

src/components/tab/SearchInput.vue
individual-it: fixed in 7855744b0bb3713738499d5a4e9ae1027d79b9fd

src/components/tab/SearchInput.vue
individual-it: good point, fixed in 38b437d60ccdb9ee223608de8e81b380624df7c5

kiranparajuli589 (commented):
lgtm :+1: needs rebase

kiranparajuli589 (approved):

PR #80, opened 2022-03-21
get OAuth config in request function and don't pass the data through the different layers

individual-it:
get configuration like token, url, etc. in the function that needs it `request` and don't pass this data through multiple levels

I think the code is more readable that way.

Thoughts?


kiranparajuli589 (approved):
making requests is now much easier, :cool:  :+1:

PR #81, opened 2022-03-21
use fit-content for workpackage status

kiranparajuli589:
## Description
currently, we're using a static width for work package status in the search result.

work package status text may have variable length, so the static width may lead to UI inconsistency like:

`very-long-status-te`~~xt~~ is missed because it does not fit inside specified static width
`bug................`  a long gap for short status

This PR makes use of dynamic width for the work package status indicator.

### DIFF
--- a/src/components/tab/WorkPackage.vue
+++ b/src/components/tab/WorkPackage.vue
@@ -67,8 +67,8 @@ export default {
 		flex-wrap: wrap;
 
 		&__status {
-			padding: 5px;
-			width: 98px;
+			padding: 5px 10px;
+			width: fit-content;
 			height: 20px;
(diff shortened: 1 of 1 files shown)


individual-it (commented):
can you please post before-after screen shots?

kiranparajuli589:
Before:
Screenshot from 2022-03-21 14-05-21
Screenshot from 2022-03-21 13-55-50

After:
Screenshot from 2022-03-21 14-07-44
Screenshot from 2022-03-21 14-07-01

> PS open project allow the status to be up to 30 characters

SwikritiT (commented):
LGTM

individual-it (approved):

PR #82, opened 2022-03-21
[op#41470] reset search input state if fileInfo prop has changed

kiranparajuli589:
### Description
Reset search input state if the `fileInfo` prop changes

### Related
[OP#41470](https://community.openproject.org/projects/nextcloud-integration/work_packages/41470)

Signed-off-by: Parajuli Kiran <kiranparajuli589@gmail.com>

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -1,6 +1,6 @@
 <template>
 	<div id="searchBar">
-		<Multiselect id="search-input"
+		<Multiselect ref="workPackageMultiSelect"
 			class="searchInput"
 			:placeholder="placeholder"
 			:options="searchResults"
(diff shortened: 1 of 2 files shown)


src/components/tab/SearchInput.vue
```diff
@@ -1,6 +1,7 @@
 <template>
 	<div id="searchBar">
-		<Multiselect class="searchInput"
+		<Multiselect id="search-input"
```
individual-it, line 3: there might be other items on the page with that generic id

individual-it, line 3: there might be other items on the page with that generic id

src/components/tab/SearchInput.vue
```diff
@@ -79,6 +80,15 @@ export default {
 			return ''
 		},
 	},
+	watch: {
+		fileInfo(oldFile, newFile) {
+			if (oldFile.id !== newFile.id) {
+				this.selectedId = []
+				this.resetState()
+				document.getElementById('search-input').value = ''
```
individual-it, line 88: this does not work for me, the search  field still shows the previously typed value

individual-it, line 88: this does not work for me, the search  field still shows the previously typed value

individual-it (changes requested):

kiranparajuli589:
@individual-it I've addressed your reviews. please have a look again.

individual-it (approved):

PR #88, opened 2022-03-22
[OP#41127] filter work packages by storage and sort by status

individual-it:
when searching by query (to link a wp to file):
- don't filter by a magic status id, but by the meta-status "open" 
- additionally filter out packages that cannot be linked to the current storage

CC @Kharonus

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -236,6 +236,7 @@ class OpenProjectAPIService {
 	 * @param string $userId
 	 * @param string|null $query
 	 * @param int|null $fileId
+	 * @param string|null $storageUrl
 	 * @param int $offset
 	 * @param int $limit
 	 * @return array<mixed>
@@ -243,8 +244,18 @@ class OpenProjectAPIService {
 	 * @throws \Safe\Exceptions\JsonException
 	 */
 	public function searchWorkPackage(
-		string $userId, string $query = null, int $fileId = null, int $offset = 0, int $limit = 5
+		string $userId,
+		string $query = null,
+		int $fileId = null,
+		string $storageUrl = null,
+		int $offset = 0,
+		int $limit = 5
 	): array {
+		$linkableStorageFilter = [
+			'linkable_to_storage_url' =>
+				['operator' => '=', 'values' => [$storageUrl]]
+		];
+		$openStatusFilter = ['status' => ['operator' => 'o', 'values' => []]];
 		$resultsById = [];
 		$filters = [];
 
@@ -254,7 +265,10 @@ class OpenProjectAPIService {
 		}
 		if ($query !== null) {
 			$filters[] = ['description' => ['operator' => '~', 'values' => [$query]]];
-			$filters[] = ['status' => ['operator' => '!', 'values' => ['14']]];
+			$filters[] = $openStatusFilter;
+			if ($storageUrl !== null) {
+				$filters[] = $linkableStorageFilter;
+			}
 		}
 		$params = [
 			'filters' => \Safe\json_encode($filters),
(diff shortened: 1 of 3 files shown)


kiranparajuli589 (approved):
lgtm :+1:

lib/Service/OpenProjectAPIService.php
```diff
@@ -236,15 +236,26 @@ public function getNotifications(
...
+		string $userId,
+		string $query = null,
+		int $fileId = null,
+		string $storageUrl = null,
+		int $offset = 0,
+		int $limit = 5
 	): array {
+		$linkableStorageFilter = [
+			'linkable_to_storage_url' =>
+				['operator' => '=', 'values' => [$storageUrl]]
```
individual-it, line 256: ToDo: urlencode

individual-it:
@Kharonus encoding the URL as requested by you. Could you please review the filters

Kharonus:
> @Kharonus encoding the URL as requested by you. Could you please review the filters

The used filters look correct from syntax and semantic. Just one question remains: Should we include now the closed tickets in the search result? And add a `sortBy` to the request?

individual-it:
@Kharonus @kiranparajuli589 I've changed it so, that its now does not filter by status at all, but only by the storage and shows results of closed status at the end.
Please review again

individual-it:
don't undestand why CI is failing, works locally also with newest nextcloud, need to investigate

individual-it:
CI fails because of this change in NC master https://github.com/nextcloud/server/commit/6be7aa112f6174c1406cfef5449836124a7f6f50#diff-cd64d65cb5d9cb2c60c6a050ac0739f4b94390011ef66ee81523885f60a4c64bL63
PR: https://github.com/nextcloud/server/pull/31609

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -150,16 +151,29 @@ private function getOpenProjectAPIService(
...
-			$this->createMock(\OC\Http\Client\LocalAddressChecker::class)
-		);
+
+		//changed from nextcloud 24
+		if (version_compare(OC_Util::getVersionString(), '24') >= 0) {
+			// @phpstan-ignore-next-line
+			$ocClient = new Client(
+				$this->createMock(IConfig::class),                             // @phpstan-ignore-line
+				$certificateManager,                                                          // @phpstan-ignore-line
+				$client,                                                                      // @phpstan-ignore-line
```
kiranparajuli589, line 163: indentations are a little too more here.

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -150,16 +151,29 @@ private function getOpenProjectAPIService(
...
+				$this->createMock(\OC\Http\Client\LocalAddressChecker::class)  // @phpstan-ignore-line
+			);
+		} else {
+			// @phpstan-ignore-next-line
+			$ocClient = new Client(
+				$this->createMock(IConfig::class),                             // @phpstan-ignore-line
+				$this->createMock(ILogger::class),                             // @phpstan-ignore-line
+				$certificateManager,                                                          // @phpstan-ignore-line
+				$client,                                                                      // @phpstan-ignore-line
```
kiranparajuli589, line 172: indentations are a little too more here

kiranparajuli589 (commented):
very minor comments

tests/lib/Service/OpenProjectAPIServiceTest.php
individual-it: moved the fix to https://github.com/nextcloud/integration_openproject/pull/95 and fixed the indent there, please review there

PR #91, opened 2022-03-28
Use debounced function to make new search request

kiranparajuli589:
### Description
With this PR: `makeSearchRequest` function is debounced 500 ms to make a new search request.

### Related
- OP#41514
- https://community.openproject.org/projects/nextcloud-integration/work_packages/41514

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -30,6 +30,7 @@
 </template>
 
 <script>
+import debounce from 'lodash/debounce'
 import axios from '@nextcloud/axios'
 import { generateUrl } from '@nextcloud/router'
 import { translate as t } from '@nextcloud/l10n'
@@ -38,13 +39,12 @@ import WorkPackage from './WorkPackage'
 import { showError } from '@nextcloud/dialogs'
 import { workpackageHelper } from '../../utils/workpackageHelper'
 
-const _ = require('lodash')
-
 const STATE_OK = 'ok'
 const STATE_ERROR = 'error'
 const STATE_NO_TOKEN = 'no-token'
 const STATE_LOADING = 'loading'
 const SEARCH_CHAR_LIMIT = 3
+const DEBOUNCE_THRESHOLD = 500
 
 export default {
 	name: 'SearchInput',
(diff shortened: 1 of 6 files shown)


src/components/tab/SearchInput.vue
```diff
@@ -110,6 +111,17 @@ export default {
 				this.state = STATE_ERROR
 			}
 		},
+		async asyncFind(query) {
+			query = query.trim()
```
individual-it, line 115: do we want to search for things like `new `?

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -159,17 +162,18 @@ describe('SearchInput.vue tests', () => {
...
 				})
-				const searchList = wrapper.find(searchListSelector)
-
-				expect(searchList.exists()).toBeTruthy()
-				expect(searchList).toMatchSnapshot()
+				const multiSelectContent = wrapper.find(multiSelectContentSelector)
+				expect(multiSelectContent.exists()).toBeTruthy()
+				const workPackages = multiSelectContent.findAll(workPackageStubSelector)
+				expect(workPackages).toHaveLength(workPackagesSearchResponse.length)
+				expect(workPackages.at(0).props()).toMatchSnapshot()
```
individual-it, line 176: why checking only one and not all?

src/components/tab/SearchInput.vue
```diff
@@ -110,6 +111,17 @@ export default {
...
+			if (query) {
+				this.resetState()
+				if (query.length <= SEARCH_CHAR_LIMIT) return
+				await this.debounceMakeSearchRequest(query)
+			}
+		},
+		debounceMakeSearchRequest: _.debounce(function(...args) {
+			return this.makeSearchRequest(...args)
+		}, 500),
```
individual-it, line 124: put the number in some const?

individual-it (changes requested):
small comments
but it seems not to work for me
image

src/components/tab/SearchInput.vue
kiranparajuli589: done

tests/jest/components/tab/SearchInput.spec.js
kiranparajuli589: fixed

src/components/tab/SearchInput.vue
kiranparajuli589: trim removed

kiranparajuli589:
> small comments but it seems not to work for me image

@individual-it it works for me, can you please check again. I've checked with node v14 and v16 and it works for both. I'm using npm 8.5.0

individual-it:
something went wrong with CI, restarted

individual-it (approved):
works now.
Any way we could catch that kind of issue with tests, so that it would not happen again?

kiranparajuli589:
> works now. Any way we could catch that kind of issue with tests, so that it would not happen again?

This was a javascript error. occ app enable/disable was working fine. I think we have to open the UI to check this. acceptance would have catched this problem.

PR #93, opened 2022-03-30
change postgresql port mapping to not conflict with openproject db

kiranparajuli589:
While using a local open project setup with PostgreSQL DB and docker-compose for nextcloud and the integration app:
```bash
foreman start -f Procfile.dev
```
exits being not happy with port `5433` for the database.
```console
Error response from daemon: driver failed programming external connectivity on endpoint integration_openproject_db (1ec90895fe355fa3bc3cbaf4c694596cde695bf9672f33ba3a9d2d8a519bc51e): Error starting userland proxy: listen tcp4 0.0.0.0:5433: bind: address already in use
```

With this PR, the integration app DB port is mapped to more unique code for the host machine.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     container_name: integration_openproject_db
     restart: always
     ports:
-      - "5433:5432"
+      - "2345:5432"
     volumes:
       - db:/var/lib/postgresql/data
     environment:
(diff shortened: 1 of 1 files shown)


kiranparajuli589:
seems like this is only appearing in my machine. for me `5432` is up while I start my computer. maybe used by some application.
if people get this error, they can change the mapped port to some unique one or kill that port before starting the docker compose.

with this infor, i think this can be closed.

PR #97, opened 2022-04-08
Automatically create a NC OAuth client for OpenProject

eneiluj:
Hey, here is a basic implementation of automatic creation of a Nextcloud OAuth client for the currently configured OpenProject instance.

Basically, in the admin settings, a button shows up if no NC OAuth client has been created yet. It triggers the creation of a NC OAuth client and the ID/secret are then displayed. When the OP instance URL changes, the client is automatically deleted.

TODO:
* [ ] The generated redirect URL is most likely invalid. It is now `https://OP_INSTANCE_URL/oauth/redirect`
* [ ] Improve admin settings design, messages and labels

@wielinde It looks like that:
op

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -20,6 +20,7 @@ use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Http\RedirectResponse;
 use OCP\AppFramework\Controller;
 
+use OCA\OpenProject\Service\OauthService;
 use OCA\OpenProject\Service\OpenProjectAPIService;
 use OCA\OpenProject\AppInfo\Application;
 use Psr\Log\LoggerInterface;
@@ -52,6 +53,10 @@ class ConfigController extends Controller {
 	 * @var string|null
 	 */
 	private $userId;
+	/**
+	 * @var OauthService
+	 */
+	private $oauthService;
 
 	public function __construct(string $appName,
 								IRequest $request,
@@ -60,6 +65,7 @@ class ConfigController extends Controller {
 								IL10N $l,
 								OpenProjectAPIService $openprojectAPIService,
 								LoggerInterface $logger,
+								OauthService $oauthService,
 								?string $userId) {
 		parent::__construct($appName, $request);
 		$this->config = $config;
@@ -68,6 +74,7 @@ class ConfigController extends Controller {
 		$this->openprojectAPIService = $openprojectAPIService;
 		$this->logger = $logger;
 		$this->userId = $userId;
+		$this->oauthService = $oauthService;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


lib/Service/OauthService.php
```diff
@@ -0,0 +1,102 @@
...
+
+	/**
+	 * @param string $name
+	 * @param string $redirectUri
+	 * @return array<mixed>
+	 */
+	public function createNcOauthClient(string $name, string $redirectUri): array {
+		$client = new Client();
+		$client->setName($name);
+		$client->setRedirectUri($redirectUri);
```
wielinde, line 55: Hey @eneiluj, would it be possible to generate the redirect URI differently? Would be great to have the following pattern `<OP HOST>/oauth_clients/<oauth client id>/callback`. So the important change is that the URL contains the identifier (the long string). Then the admin would not need to have to copy another identifier (as there can be multiple Nextcloud instances connected to the OP instance).

lib/Controller/ConfigController.php
```diff
@@ -246,4 +261,23 @@ private function storeUserInfo(): array {
...
+		if ($oauthClientInternalId !== '') {
+			$id = (int) $oauthClientInternalId;
+			$clientInfo = $this->oauthService->getClientInfo($id);
+			if ($clientInfo !== null) {
+				return new DataResponse($clientInfo);
+			}
+		}
+
+		$opUrl = $this->config->getAppValue(Application::APP_ID, 'oauth_instance_url', '');
+		$clientInfo = $this->oauthService->createNcOauthClient('OpenProject client', rtrim($opUrl, '/') .'/oauth_clients/%s/callback');
```
SwikritiT, line 279: @wielinde the `oauth` redirect URI follows the pattern `<OP HOST>/oauth_clients/<oauth client id>/callback`  and the client id is added here:
https://github.com/nextcloud/integration_openproject/blob/eeb4a6f9ce0f6eeda8103b01b31622a328bea7a2/lib/Service/OauthService.php#L52-L62

Does this look good?

lib/Controller/ConfigController.php
```diff
@@ -137,6 +144,14 @@ public function setAdminConfig(array $values): DataResponse {
 		$this->userManager->callForAllUsers(function (IUser $user) {
 			$this->clearUserInfo($user->getUID());
 		});
+		if (isset($values['oauth_instance_url'])) {
```
kiranparajuli589, line 147: if the OpenProject OAuth instance URL is changed and we already have an NC OAuth client created for OP, the NC OAuth client is deleted and the stored `nc_oauth_client_id` is also cleared. :+1:

kiranparajuli589:
new tests and features to reset NC OAuth Client can be added in different PRs. @SwikritiT @individual-it it will be helpful for the redesign PR if we can merge this.

individual-it (approved):
let's merge this now as it is, there will be some heavy refactoring to be done in #124

PR #102, opened 2022-04-28
[OP#41504]refactor  OpenProjectAPIService::getOpenProjectAvatar so it uses  Ope…

SwikritiT:
This PR refactors OpenProjectAPIService::getOpenProjectAvatar so it uses OpenProjectAPIService::request
OP#41504 : https://community.openproject.org/projects/nextcloud-integration/work_packages/41504/activity

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -19,6 +19,7 @@ use OCA\OpenProject\Exception\OpenprojectErrorException;
 use OCA\OpenProject\Exception\OpenprojectResponseException;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
+use OCP\Http\Client\IResponse;
 use OCP\IL10N;
 use OCP\IURLGenerator;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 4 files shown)


kiranparajuli589 (commented):
lgtm :+1:

lib/Service/OpenProjectAPIService.php
```diff
@@ -364,6 +354,62 @@ public function getOpenProjectAvatar(string $url,
...
 	}
 
+	/**
+	 * @param string $accessToken
+	 * @param string $openprojectUrl
+	 * @param string $endPoint
+	 * @param array<mixed> $params
+	 * @param string $method
+	 */
+	// @phpstan-ignore-next-line
```
individual-it, line 364: what do we want to ignore here?

lib/Service/OpenProjectAPIService.php
```diff
@@ -364,6 +354,62 @@ public function getOpenProjectAvatar(string $url,
...
+	public function rawRequest(string $accessToken, string $openprojectUrl,
```
individual-it, line 365: can be private as its only used inside of this class

individual-it (changes requested):

individual-it (approved):

PR #104, opened 2022-05-04
OP#41949 allways send storageUrl

individual-it:
the search and the link function could be used without or a wrong storageUrl, this is actually not needed, I'm suggesting in this PR to send the storageURL in every link and search request to OpenProject

should fix https://community.openproject.org/work_packages/41949 
In the search for listing existing  links we did not send the storageUrl as filter so also links from other storages would show up


kiranparajuli589 (approved):
lgtm :+1:

PR #108, opened 2022-05-05
OP#41271 remember new refresh_token after refreshing oauth token

individual-it:
part of https://community.openproject.org/projects/nextcloud-integration/work_packages/41271

when refreshing the oauth token, open project not only sends back a new token but also a new refresh-token, that also needs to be remembered, otherwise the second refresh would fail.

CC @Kharonus


kiranparajuli589 (approved):
lgtm :+1:

PR #110, opened 2022-05-06
add confirm dialog for admin config update action

kiranparajuli589:
#### Description
- implement confirm dialog to update admin credentials
- delete user prefs. if updated admin config is not ok
- added button to submit the form

#### Related
- [OP#42273] part of https://community.openproject.org/projects/nextcloud-integration/work_packages/42273

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -134,18 +134,12 @@ class ConfigController extends Controller {
 		foreach ($values as $key => $value) {
 			$this->config->setAppValue(Application::APP_ID, $key, trim($value));
 		}
-		$configStatus = OpenProjectAPIService::isAdminConfigOk($this->config);
 		// if the admin config is not ok, we clear every existing user information
-		if (!$configStatus) {
-			$this->userManager->callForAllUsers(function (IUser $user) {
-				$this->clearUserInfo($user->getUID());
-			});
-			return new DataResponse([
-				"status" => false
-			]);
-		}
+		$this->userManager->callForAllUsers(function (IUser $user) {
+			$this->clearUserInfo($user->getUID());
+		});
 		return new DataResponse([
-			"status" => true
+			"status" => OpenProjectAPIService::isAdminConfigOk($this->config)
 		]);
 	}
 
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -82,7 +82,7 @@ class ConfigController extends Controller {
 	 * @param string|null $userId
 	 * @return void
 	 */
-	private function clearUserInfo(string $userId = null) {
+	public function clearUserInfo(string $userId = null) {
 		if ($userId === null) {
 			$userId = $this->userId;
 		}
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -15,6 +15,8 @@ use OCP\IURLGenerator;
 use OCP\IConfig;
 use OCP\IL10N;
 use OCP\IRequest;
+use OCP\IUser;
+use OCP\IUserManager;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Http\RedirectResponse;
@@ -34,6 +36,10 @@ class ConfigController extends Controller {
 	 * @var IURLGenerator
 	 */
 	private $urlGenerator;
+	/**
+	 * @var IUserManager
+	 */
+	private $userManager;
 	/**
 	 * @var IL10N
 	 */
@@ -57,6 +63,7 @@ class ConfigController extends Controller {
 								IRequest $request,
 								IConfig $config,
 								IURLGenerator $urlGenerator,
+								IUserManager $userManager,
 								IL10N $l,
 								OpenProjectAPIService $openprojectAPIService,
 								LoggerInterface $logger,
@@ -64,10 +71,27 @@ class ConfigController extends Controller {
 		parent::__construct($appName, $request);
 		$this->config = $config;
 		$this->urlGenerator = $urlGenerator;
+		$this->userManager = $userManager;
 		$this->l = $l;
 		$this->openprojectAPIService = $openprojectAPIService;
 		$this->logger = $logger;
 		$this->userId = $userId;
+	}
+
+	/**
+	 * @param string|null $userId
+	 * @return void
+	 */
+	private function clearUserInfo(string $userId = null) {
+		if ($userId === null) {
+			$userId = $this->userId;
+		}
+		$this->config->deleteUserValue($userId, Application::APP_ID, 'token');
+		$this->config->deleteUserValue($userId, Application::APP_ID, 'login');
+		$this->config->deleteUserValue($userId, Application::APP_ID, 'user_id');
+		$this->config->deleteUserValue($userId, Application::APP_ID, 'user_name');
+		$this->config->deleteUserValue($userId, Application::APP_ID, 'refresh_token');
+		$this->config->deleteUserValue($userId, Application::APP_ID, 'last_notification_check');
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


lib/Controller/ConfigController.php
```diff
@@ -115,7 +134,13 @@ public function setAdminConfig(array $values): DataResponse {
 		foreach ($values as $key => $value) {
 			$this->config->setAppValue(Application::APP_ID, $key, trim($value));
 		}
-		return new DataResponse(1);
+		// if the admin config is not ok, we clear every existing user information
```
individual-it, line 137: we clear it in any case

individual-it, line 137: we clear it in any case

src/components/AdminSettings.vue
```diff
@@ -65,17 +68,35 @@ export default {
...
+		updateForm() {
 			const that = this
-			delay(() => {
-				that.saveOptions()
-			}, 2000)()
+			OC.dialogs.confirmDestructive(
+				t(
+					'integration_openproject',
+					'Are you sure you want to update the admin settings?'
+					+ ' After saving, every connected users must need to re-connect to the Openproject instance.'
```
individual-it, line 82:
```suggestion
					+ ' After saving, every connected user will have to re-connect to the OpenProject instance.'
```

src/components/AdminSettings.vue
```diff
@@ -65,17 +68,35 @@ export default {
...
-		onInput() {
+		updateForm() {
 			const that = this
-			delay(() => {
-				that.saveOptions()
-			}, 2000)()
+			OC.dialogs.confirmDestructive(
+				t(
+					'integration_openproject',
+					'Are you sure you want to update the admin settings?'
```
individual-it, line 81:
```suggestion
					'Are you sure you want to update the Oauth settings for OpenProject?'
```

tests/jest/components/AdminSettings.spec.js
```diff
@@ -40,29 +61,99 @@ describe('AdminSettings', () => {
...
+				confirmSpy = jest.spyOn(global.OC.dialogs, 'confirmDestructive')
+				wrapper = getWrapper({
+					isAdminConfigOk: false,
 				})
-			const wrapper = getWrapper()
-			const inputField = wrapper.find(inputSelector)
-
-			await inputField.trigger('input')
+			})
+			it('should show the save button', async () => {
```
individual-it, line 76:
```suggestion
			it('should show the update button', async () => {
```

tests/jest/components/AdminSettings.spec.js
```diff
@@ -40,29 +61,99 @@ describe('AdminSettings', () => {
...
+				expect(wrapper.find(selectors.saveConfigButton)).toMatchSnapshot()
+				expect(wrapper.find(selectors.updateConfigButton).exists()).toBeFalsy()
```
individual-it, line 78: I think this should be true

tests/jest/components/AdminSettings.spec.js
```diff
@@ -40,29 +61,99 @@ describe('AdminSettings', () => {
...
 			jest.clearAllMocks()
 		})
-		it.each([
-			selectors.oauthInstance,
-			selectors.oauthClientId,
-			selectors.oauthClientId,
-		])('should call "onInput" method', async (inputSelector) => {
-			const onInputSpy = jest.spyOn(AdminSettings.methods, 'onInput')
-				.mockImplementationOnce(() => {
+		describe('when the admin config status is not ok beforehand', () => {
```
individual-it, line 68:
```suggestion
		describe('admin config is not complete', () => {
```

tests/jest/components/AdminSettings.spec.js
```diff
@@ -40,29 +61,99 @@ describe('AdminSettings', () => {
...
+			})
+			it('should not trigger confirm dialog on save', async () => {
+				const saveConfigButton = wrapper.find(selectors.saveConfigButton)
+				const inputField = wrapper.find(selectors.oauthClientId)
+				await inputField.setValue('test')
+				await saveConfigButton.trigger('click')
+				expect(confirmSpy).toBeCalledTimes(0)
+			})
+		})
+		describe('when the admin config status is ok beforehand', () => {
```
individual-it, line 88:
```suggestion
		describe('admin config is complete', () => {
```

tests/jest/components/AdminSettings.spec.js
```diff
@@ -40,29 +61,99 @@ describe('AdminSettings', () => {
...
+			let wrapper, confirmSpy
+			beforeEach(() => {
+				confirmSpy = jest.spyOn(global.OC.dialogs, 'confirmDestructive')
+				wrapper = getWrapper({
+					isAdminConfigOk: true,
+				})
+			})
+			it('should show the save button', async () => {
+				expect(wrapper.find(selectors.updateConfigButton)).toMatchSnapshot()
+				expect(wrapper.find(selectors.saveConfigButton).exists()).toBeFalsy()
```
individual-it, line 98: shouldn't that be true?

tests/lib/Controller/ConfigControllerTest.php
```diff
@@ -286,4 +298,107 @@ public function testOauthNoAccessTokenInResponse($oauthResponse, $expectedRedire
...
+	}
+
+	/**
+	 * @param array<string> $credsToUpdate
+	 * @param bool $adminConfigStatus
+	 * @dataProvider setAdminConfigStatusDataProvider
+	 * @return void
+	 */
+	public function testSetAdminConfigForDifferentAdminConfigStatus($credsToUpdate, $adminConfigStatus) {
+		$manager = \OC::$server->getUserManager();
```
individual-it, line 332:
```suggestion
		$userManager = \OC::$server->getUserManager();
```

individual-it (changes requested):

tests/lib/Controller/ConfigControllerTest.php
```diff
@@ -286,4 +298,107 @@ public function testOauthNoAccessTokenInResponse($oauthResponse, $expectedRedire
...
+		$count = 0;
+		$function = function (IUser $user) use (&$count) {
+			$count++;
+		};
+		$manager->callForAllUsers($function, '', true);
+		$this->assertSame(1, $count, 'Expected to have only 1 user in the dB before this test');
```
individual-it, line 338: why should we fail tests if there are more then one user before?

kiranparajuli589: I've to mock the `deleteUserValue` fn for the number of fn calls and consecutive args to verify cleanup is done for every user. The consecutive args `withConsecutive` can't be fed calculated data, so I have to have only two to match with the CI. One will be an admin and another will be the test-created user.

tests/jest/components/AdminSettings.spec.js
kiranparajuli589: when the admin config is incomplete, it shows the save btn without the confirm dialog. isn't that how it should behave?

tests/jest/components/AdminSettings.spec.js
kiranparajuli589: if the admin config is incomplete, only the save btn is displayed

tests/jest/components/AdminSettings.spec.js
kiranparajuli589: if the admin config is complete, then only update btn is displayed which then triggers the confirm dialog. isn't this how it should be?

tests/lib/Controller/ConfigControllerTest.php
kiranparajuli589: if we can provide the args like this then the check wouldn't have been required. can we do it like so?
```
->withConsecutive(listOfConsecutiveArgs)
```

src/components/AdminSettings.vue
```diff
@@ -65,17 +68,35 @@ export default {
...
+					'Are you sure you want to update the Oauth settings for OpenProject?'
+					+ ' After saving, all connected users will have to re-connect to the OpenProject instance.'
+				),
+				t('integration_openproject', 'Confirm Update'),
+				{
+					type: OC.dialogs.YES_NO_BUTTONS,
+					confirm: t('integration_openproject', 'Update'),
+					confirmClasses: 'error',
+					cancel: t('integration_openproject', 'Cancel'),
+				},
```
individual-it, line 90: replace with suggestion of @wielinde in https://community.openproject.org/projects/nextcloud-integration/work_packages/42273/activity#activity-4

individual-it (changes requested):

individual-it (approved):

individual-it:
this should also fix https://community.openproject.org/projects/nextcloud-integration/work_packages/42272 or?

kiranparajuli589:
> this should also fix https://community.openproject.org/projects/nextcloud-integration/work_packages/42272 or?

yes it should

PR #111, opened 2022-05-09
[OP#41678] use notifications API to fetch OP notifications

individual-it:
use the notification API to fetch notifications from OpenProject (for dashboard and for the NC notifications)

please note for the notifications to work you need to install and enable the notifications NC app https://github.com/nextcloud/notifications and also enable `Enable notifications for activity in my work packages` in the openproject_integration personal settings

dashboard view
image

notifications view
image

The count of notifications in the notifications view is the count of NEW notifications since the last check. Dismissing notifications on NC side will NOT have any effect on the OP notifications

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/41678

### DIFF
--- a/lib/Controller/OpenProjectAPIController.php
+++ b/lib/Controller/OpenProjectAPIController.php
@@ -110,14 +110,13 @@ class OpenProjectAPIController extends Controller {
 	 * get notifications list
 	 * @NoAdminRequired
 	 *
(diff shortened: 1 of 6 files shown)


kiranparajuli589 (approved):
lgtm :+1:

PR #112, opened 2022-05-11
Fix the misplaced route

SwikritiT:
The endpoint for `deleting` a link should be inside `routes` but PR https://github.com/nextcloud/integration_openproject/pull/67/files#diff-6301937af1cc1575e9e18ecd27c4a51f0d3cf9ce12e59a14709f626dfa3ef952 misplaced it 

This PR fixes that. This fixes the unlinking of the work package which wasn't working before.


individual-it (approved):
prove that we are lacking e2e tests

PR #113, opened 2022-05-13
OP#41503 only show WP belonging to this file (async cases)

individual-it:
switching between files while the list is still loading might add data from an other file to the current view, see https://community.openproject.org/projects/nextcloud-integration/work_packages/41503

To protect from that this PR:
1. checks is a WP already exists in the list (by comparing the fileId and the WPid)
2. filters the workpackages to show by the fileId of the file currently in view

manual test-cases
| test | expected outcome  | result |
| --- | --- | --- |
| 1. link file1 to WP-A, WP-B, WP-C 2. link file2 to WP-D, WP-E 3. quickly switch between the files | WP list is correct, no extra WP listed | :heavy_check_mark: |
| 1. link file1 to WP-A, WP-B, WP-C 2. link file2 to WP-C, WP-D 3. quickly switch between the files | WP list is correct, no extra WP listed, WP-C is shown in both files | :heavy_check_mark: |
| 1. link file1 to WP-A, WP-B, WP-C 2. don't link file2 to any WP 3. quickly switch between the files, ending on file2 | WP list is empty, no links message shown | :heavy_check_mark: |
| 1. link file1 to WP-A, WP-B, WP-C 2. link file2 to WP-D, WP-E 3. quickly switch between the files | WP order is the same as in the case when the user would have waited for the complete loading  | :x: |

ToDo:
- [ ] ~~hardening existing unit tests (e.g. without `workPackage.fileId = this.fileInfo.id` in `processWorkPackages` adding WP would not work, but no test failed)~~ integration or better e2e tests needed for that
- [ ] ~~write tests for the cases above~~ I couldn't achieve it, maybe a case for manual testing
- [ ] discuss if we do some "local" sorting for the last case

@kiranparajuli589 @SwikritiT @eneiluj putting this to draft for now, but happy to hear your input

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -169,6 +169,7 @@ export default {
 			for (let workPackage of workPackages) {
 				try {
 					if (this.isStateLoading) {
+						workPackage.fileId = this.fileInfo.id
 						workPackage = await workpackageHelper.getAdditionalMetaData(workPackage)
 						const alreadyLinked = this.linkedWorkPackages.some(el => el.id === workPackage.id)
 						const alreadyInSearchResults = this.searchResults.some(el => el.id === workPackage.id)
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -3,7 +3,7 @@
 		<Multiselect ref="workPackageMultiSelect"
 			class="searchInput"
 			:placeholder="placeholder"
-			:options="searchResults"
+			:options="filterSearchResultsByFileId"
 			:user-select="true"
 			label="displayName"
 			track-by="id"
@@ -81,6 +81,15 @@ export default {
 			}
 			return ''
 		},
+		filterSearchResultsByFileId() {
+			return this.searchResults.filter(wp => {
+				if (wp.fileId === undefined || wp.fileId === '') {
+					console.error('work-package data does not contain a fileId')
+					return false
+				}
+				return wp.fileId === this.fileInfo.id
+			})
+		},
 	},
 	watch: {
 		fileInfo(oldFile, newFile) {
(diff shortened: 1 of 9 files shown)


individual-it:
search result can have a similar issue if the user is fast enough
1. type into the search-field
2. while the search is running switch to an other file
3. click into the search field of the newly selected file
4. the search of the former file would be listed

This issue should also be fixed with this PR

src/components/tab/SearchInput.vue
```diff
@@ -159,12 +168,13 @@ export default {
...
 			this.checkForErrorCode(response.status)
-			if (response.status === 200) await this.processWorkPackages(response.data)
+			if (response.status === 200) await this.processWorkPackages(response.data, fileId)
 			if (this.isStateLoading) this.state = STATE.OK
 		},
-		async processWorkPackages(workPackages) {
+		async processWorkPackages(workPackages, fileId) {
 			for (let workPackage of workPackages) {
 				try {
+					workPackage.fileId = fileId
```
kiranparajuli589, line 177: `this.fileInfo.id` will be the same for whole `searchInput`. you probably can use it like this without having to use it as a fn arg.
```suggestion
					workPackage.fileId = this.fileInfo.id
```

individual-it: `this.fileInfo.id` changes whenever an other file is selected in the UI, that is the basic problem. Here we want to save the fileId of the time when the search was started, so that the found workpackage will be associated with that particular fileId even when the user selects an other file (and `this.fileInfo.id` changes)

kiranparajuli589 (approved):

individual-it:
merging this for now, for the last thing I've opened a new WP https://community.openproject.org/projects/nextcloud-integration/work_packages/42612

PR #114, opened 2022-05-17
OP#42140 only addd to searchlist when in loading state

individual-it:
fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/42140

when a link gets created the state switches from `loading` to `ok` and the list of workpackages in the search gets cleaned, but if some requests were outstanding workpackages would be kept on added to the list and it would not be cleaned correctly.

This PR makes sure workpackages are not processed and the searchlist is not extended while the state is anything else than `loading`

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -165,11 +165,14 @@ export default {
 		async processWorkPackages(workPackages) {
 			for (let workPackage of workPackages) {
 				try {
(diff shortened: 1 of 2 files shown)


individual-it:
also fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/41517
and partly https://community.openproject.org/projects/nextcloud-integration/work_packages/41503

kiranparajuli589 (approved):

PR #115, opened 2022-05-17
Do not show search input unless the projects tab is completely loaded

kiranparajuli589:
### Description
We check for the request-url, before showing the search input the users. If it is not valid, it is kept hidden, but if the url is valid, the input is shown early even if the projects tab is still loading fetching for the linked work packages.

With this PR, the search input is shown only if the the request URL is valid and the projects tab state is not loading.

### Related
https://community.openproject.org/projects/nextcloud-integration/work_packages/41496

### DIFF
--- a/src/views/ProjectsTab.vue
+++ b/src/views/ProjectsTab.vue
@@ -22,7 +22,7 @@
 
 <template>
 	<div class="projects">
-		<SearchInput v-if="!!requestUrl"
+		<SearchInput v-if="!!requestUrl && !isLoading"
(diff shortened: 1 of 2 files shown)


individual-it:
will this fix https://community.openproject.org/projects/nextcloud-integration/work_packages/41496 ?

kiranparajuli589:
> will this fix https://community.openproject.org/projects/nextcloud-integration/work_packages/41496 ?

yes it should fix OP#41496

individual-it (approved):

PR #122, opened 2022-05-31
[#42634] do not use speaker icon for the notification items in the dashboard widget

kiranparajuli589:
### Description
- do not use the speaker icon at all for the notification items in the dashboard widget

in the widget item component, there is a check for the icon URL if it's truthy. Ref: https://github.com/nextcloud/nextcloud-vue-dashboard/blob/fd9246f2e3102dd0b33ac1e90975a42f518354b6/src/DashboardWidgetItem.vue#L39

we set is as `false` so that it is never rendered on the browser

### Related
- https://community.openproject.org/work_packages/42634


individual-it (approved):
tested on NC 24 & 25 dev and it works fine!

PR #123, opened 2022-06-01
highlight & focus the OP instance input on invalid input

individual-it:
if the openproject URL is not valid the input field gets marked red and receives the focus
image

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/42727


kiranparajuli589 (commented):
lgtm :+1:

kiranparajuli589 (approved):

PR #129, opened 2022-06-06
OP#42613 cache colors of statuses and types

individual-it:
needs memory caching to be configured in NC, see https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/caching_configuration.html

In case local or distributed cache is available the code will cache the responses for the status & type requests for 1h

@eneiluj I would need your suggestion here. Is that the way the caching should be implemented? I tried with APCu, memcached and Redis in those cases it seems to work

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -20,6 +20,8 @@ use OCA\OpenProject\Exception\OpenprojectResponseException;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 use OCP\Http\Client\IResponse;
+use OCP\ICache;
+use OCP\ICacheFactory;
 use OCP\IL10N;
 use OCP\IURLGenerator;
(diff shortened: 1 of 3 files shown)


individual-it:
I think we can simplify the whole implementation quite a bit, because `$cacheFactory->createDistributed()` seems always to return something useful.
If no cache is configured it will return a `NullCache` object https://github.com/nextcloud/server/blob/7d272c54d013538746d6731097ec37f360effb5d/lib/private/Memcache/NullCache.php#L29
If only local cache is configured it will return a `APCu` object
If a distributed cache is configured, or distributed & local it will return the object of the distributed cache

eneiluj:
@individual-it Exactly, for the record, this is where the fallback cache provider is selected: https://github.com/nextcloud/server/blob/7d272c54d013538746d6731097ec37f360effb5d/lib/private/Memcache/Factory.php#L84

eneiluj (approved):
:+1:

PR #134, opened 2022-06-08
[OP#42805] Fix problem with search list not clearing

SwikritiT:
OP#42805 : https://community.openproject.org/projects/nextcloud-integration/work_packages/42805/activity


individual-it:
I think the issue was that when deleting quickly (holding down backspace key) the debounce function would be called on every char, but when the string become less than 4 char it would not be called at all, so after the debounce timeout expired the last call would be executed (the call with 4 char)
 
Now the debounce function is called every time and checks internally if the search limit is reached. In the case of a quick delete only the last call will be executed (after debounce timeout) and if that has a query shorter than 4 char it will return empty

see https://docs-lodash.com/v4/debounce/

individual-it (approved):

PR #136, opened 2022-06-09
New oauth service method to set the redirect uri

eneiluj:
@individual-it This does not change any other value of the OAuth client.


individual-it (approved):
CC @kiranparajuli589

kiranparajuli589:
:fireworks:

PR #137, opened 2022-06-10
OP#42613 cache colors of types and statuses in the frontedn

individual-it:
additionally to #129 this caches the colors also in JS, so that if a color was already used it even does not need to be requested from the caching server
the cache will be invalidated on page reload, but not when moving into an other folder

### DIFF
--- a/tests/jest/views/ProjectsTab.spec.js
+++ b/tests/jest/views/ProjectsTab.spec.js
@@ -6,6 +6,7 @@ import axios from '@nextcloud/axios'
 import * as initialState from '@nextcloud/initial-state'
 import { STATE } from '../../../src/utils'
 import workPackagesSearchResponse from '../fixtures/workPackagesSearchResponse.json'
+import { workpackageHelper } from '../../../src/utils/workpackageHelper'
 
 jest.mock('@nextcloud/dialogs')
 jest.mock('@nextcloud/l10n', () => ({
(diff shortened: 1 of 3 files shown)


tests/jest/views/ProjectsTab.spec.js
```diff
@@ -289,6 +295,12 @@ describe('ProjectsTab.vue Test', () => {
 				.mockImplementationOnce(() => Promise.resolve(
 					{ status: 200, data: testCase.typeColor })
 				)
+				.mockImplementationOnce(() => Promise.resolve(
+					{ status: 200, data: testCase.statusColor })
+				)
+				.mockImplementationOnce(() => Promise.resolve(
+					{ status: 200, data: testCase.typeColor })
+				)
```
individual-it, line 303: there are different statuses and different types for every WP, so the cache is not used in this case. Actually the test should not have worked in the first place and only worked because we did not call `axiosGetSpy.mockRestore()` previously so some mock remained from other tests

tests/jest/views/__snapshots__/ProjectsTab.spec.js.snap
```diff
@@ -108,7 +108,7 @@ exports[`ProjectsTab.vue Test fetchWorkpackages shows the linked workpackages 1`
     <div class="linked-workpackages--workpackage">
       <div class="workpackage linked-workpackages--workpackage--item" id="workpackage-589">
         <div class="row">
-          <div class="row__status">
+          <div class="row__status" style="background-color: rgb(165, 216, 255);">
```
individual-it, line 111: as the colors now are returned correctly by the fixed test, the snapshot needed update

kiranparajuli589 (commented):
lgtm :+1:

kiranparajuli589 (approved):

PR #142, opened 2022-06-14
Use state variable for the save options request payload

kiranparajuli589:
state variable should be used rather than the edit variable for the `saveOptions` fn.
was accidentally changed while admin settings refactoring


SwikritiT (approved):
LGTM

PR #145, opened 2022-06-16
House keeping :grin:

eneiluj:
* bump js libs
* bring back eslint/stylelint at compilation
* clarify/fix translations
    * the source parser which extracts the translations needs to find `t('APP_ID', 'TEXT')`
    * escaping can be avoided, I suggest translating link text/title independently and then inject the whole html link in the sentence
* add missing translations


src/components/AdminSettings.vue
```diff
@@ -3,12 +3,12 @@
...
 		<div class="openproject-server-host">
 			<FormHeading index="1"
-				title="OpenProject Server"
+				:title="t('integration_openproject', 'OpenProject Server')"
 				:is-complete="isServerHostFormComplete" />
 			<FieldValue v-if="isServerHostFormInView"
 				is-required
 				class="pb-1"
-				title="OpenProject host"
+				:title="t('integration_openproject', 'OpenProject host')"
```
kiranparajuli589, line 11: the prop is translated inside the component `FormHeading` https://github.com/nextcloud/integration_openproject/blob/6402685892a2698bcd2d4d68ae23c05e4bb4b105/src/components/admin/FieldValue.vue#L4

doesn't the translation work that way?

individual-it (commented):
I'm surprised by the high JS code coverage reported, e.g. we don't have a single test for the dashboard but it shows 100% coverage

src/components/AdminSettings.vue
eneiluj: Not exactly:
* The script which parses the sources to extract the translatable string must find stuff like `t('APP_ID', 'ACTUAL_STRING_AND_NOT_A_VARIABLE')` so you can't use your `translate` wrapper function.
* A component should translate what's its responsibility, not the strings passed as props (which are translated where they are defined).

eneiluj:
@individual-it Weird indeed. No idea why. Is it new or has it always been like that?

@kiranparajuli589 Merged with master and fixed conflicts.

PR #146, opened 2022-06-17
move generating OP OAuth URI to the backend

individual-it:
creating the PKCE code verifier and challenge in the frontend does not work reliable because the `crypto.subtle` only works in secure context see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle

So in this PR I moved generating the state and the PKCE values into the backend that returns the complete URL to the frontend

should fix https://community.openproject.org/projects/nextcloud-integration/work_packages/details/42915


individual-it:
CC @Kharonus

Kharonus:
LGTM, just as we discussed. I like it, that the frontend component has now less functionality.

kiranparajuli589 (approved):

PR #148, opened 2022-06-17
Debug why JS coverage is shown so high in #145

individual-it:


individual-it:
closing in favor of #158

PR #149, opened 2022-06-20
[OP#42922] only show copy button when copy to clipboard is possible

individual-it:
copy to clipboard is only possible in secure context, see https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
so in not secure contexts the user will have to copy manually

fixes, or avoids the problem of https://community.openproject.org/projects/nextcloud-integration/work_packages/42922


kiranparajuli589 (approved):
looks good :+1:

PR #152, opened 2022-06-20
[#OP42916] Remove requestUrl from front-end

SwikritiT:
## Description
after https://github.com/nextcloud/integration_openproject/pull/146 there is no need anymore to have the requestUrl  in the front end. This PR removes the `requestUrl` from the front end that we used, to determine if we should show the connect button to the user or not. Now instead of `requestUrl` we rely on `adminConfigStatus`

## Related workPackage
[#OP42916]  : https://community.openproject.org/projects/nextcloud-integration/work_packages/42916/activity

### DIFF
--- a/lib/Settings/Personal.php
+++ b/lib/Settings/Personal.php
@@ -60,10 +60,10 @@ class Personal implements ISettings {
 		];
 
 		try {
-			$requestUrl = OpenProjectAPIService::getOpenProjectOauthURL($this->config, $this->url);
-			$userConfig['request_url'] = $requestUrl;
+			$adminConfigStatus = OpenProjectAPIService::isAdminConfigOk($this->config);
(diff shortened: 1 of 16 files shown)


lib/Settings/Personal.php
```diff
@@ -60,10 +54,10 @@ public function getForm(): TemplateResponse {
 		];
 
 		try {
-			$requestUrl = OpenProjectAPIService::getOpenProjectOauthURL($this->config, $this->url);
-			$userConfig['request_url'] = $requestUrl;
+			$adminConfigStatus = OpenProjectAPIService::isAdminConfigOk($this->config);
```
kiranparajuli589, line 57: try catch is not necessary for this `isAdminConfigOk` fn, because it is not expected to throw any known exceptions.

individual-it (approved):
good more lines got deleted than added

PR #155, opened 2022-06-22
[OP#42909] Use material icon

SwikritiT:
## Description
The icon classes used in the Nextcloud app will be dropped in the future. So, this PR makes use of vue material icon https://github.com/robcresswell/vue-material-design-icons

## Related Workpackage
OP#42909 https://community.openproject.org/work_packages/42909/activity


src/components/tab/EmptyContent.vue
```diff
@@ -2,9 +2,9 @@
 	<div class="empty-content">
 		<div class="empty-content--wrapper">
 			<div class="empty-content--icon">
-				<img v-if="!!isAdminConfigOk && !isStateOk" :src="noConnectionSvg" alt="no connection">
-				<img v-else-if="!!isAdminConfigOk && isStateOk" :src="addLinkSvg" alt="add work package">
-				<img v-else :src="noConnectionSvg" alt="error">
+				<LinkOffIcon v-if="!!requestUrl && !isStateOk" :size="60" />
+				<LinkPlusIcon v-else-if="!!requestUrl && isStateOk" :size="60" />
+				<LinkOffIcon v-else :size="60" />
```
kiranparajuli589, line 7: imo, we should be using `isAdminConfigOk` check here

SwikritiT: done in the latest commit!

src/views/Dashboard.vue
```diff
@@ -4,8 +4,11 @@
 		:show-more-text="title"
 		:loading="isLoading">
 		<template #empty-content>
-			<EmptyContent v-if="emptyContentMessage"
-				:icon="emptyContentIcon">
+			<EmptyContent v-if="emptyContentMessage">
+				<template #icon>
+					<CheckBoldIcon v-if="state === 'ok'" :size="70" />
```
kiranparajuli589, line 9:
```suggestion
					<CheckBoldIcon v-if="state === STATE.OK" :size="70" />
```
please use the state variables from the STATE obj

src/components/PersonalSettings.vue
```diff
@@ -161,6 +164,18 @@ export default {
 	}
 }
 
+.openproject-connected {
+	display: flex;
+}
+
+.openproject-connected-checkmark {
+	padding-right: 8px;
```
eneiluj, line 172:
```suggestion
	padding-right: 8px;
	color: var(--color-success);
```

src/components/PersonalSettings.vue
```diff
@@ -13,12 +13,12 @@
 			<br><br>
 			<div v-if="connected" class="openproject-grid-form">
 				<label class="openproject-connected">
-					<a class="icon icon-checkmark-color" />
+					<CheckIcon class="openproject-connected-checkmark" :size="20" fill-color="#007500" />
```
eneiluj, line 16: You can set a green color with the "color" css attribute on material icons. There is a `--color-success` CSS variable provided by Nextcloud.

```suggestion
					<CheckIcon class="openproject-connected-checkmark" :size="20" />
```

eneiluj (changes requested):
I pushed a couple of commits to fix stuff in the `.complete` element of FormHeading:
* use the color-success variable 
* fix the icon alignment

In general, colors provided by Nextcloud global CSS are preferred for green/red/orange referring to success/error/warning.

Other than that, looks good to me. I fully reviewed the code but didn't try everything.

A few inline comments.

src/views/Dashboard.vue
SwikritiT: Done in latest push!

src/components/PersonalSettings.vue
SwikritiT: Done in the latest commit. Thank you for the suggestion!

PR #159, opened 2022-06-28
Public direct download page

eneiluj:
OpenProject generates Nextcloud [direct download](https://docs.nextcloud.com/server/22/developer_manual/client_apis/OCS/ocs-api-overview.html#direct-download) links to let users get files linked with a work package.

There is an issue with this direct download endpoint in Nextcloud. It messes with cookies and refuses to be accessed when put in a link in OpenProject.

This is a workaround. This public page takes the download token and an arbitrary file name and shows a link to the direct download endpoint. For some reason it works fine.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -15,6 +15,7 @@ return [
 		['name' => 'config#setConfig', 'url' => '/config', 'verb' => 'PUT'],
 		['name' => 'config#setAdminConfig', 'url' => '/admin-config', 'verb' => 'PUT'],
 		['name' => 'config#autoOauthCreation', 'url' => '/nc-oauth', 'verb' => 'POST'],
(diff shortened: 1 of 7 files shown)


individual-it:
Is that supposed to be just an API endpoint or also used from the UI?

P.S. I believe the CI failures are unrelated, I can try to fix them

individual-it:
Sorry I miss-understood it, this is not to generate the URL, but its the page itself that displays the download button.

Would it also work if this would do a redirect, so that the user would have one less click to do?

eneiluj:
@individual-it Yes it would probably work. I will try it in pseudo-real conditions to see if it does not mess with browsers security restrictions.

individual-it:
Isn't that also the case on the normal direct download page?

individual-it (approved):

PR #160, opened 2022-06-29
stop processing metadata on file change

individual-it:
If the user has selected an other file, stop processing the workpackages metadata


eneiluj (approved):
lgtm but i didn't try it.

PR #161, opened 2022-06-30
sanitize hint text

individual-it:
this is currently not strictly needed because the hint text is compiled without any user-input. The only thing that is taken from outside is the URL.
Still I would like to sanitize it here because just ignoring the message might get us into trouble one day if the way how we generate the message is changed one day and it goes unnoticed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
 	"requires": true,
 	"packages": {
 		"": {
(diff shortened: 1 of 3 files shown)


eneiluj (approved):
Yep

kiranparajuli589 (approved):

PR #163, opened 2022-07-01
Add custom direct download page actually serving the file

eneiluj:
Better alternative to the other public page which shows a link. This one checks the direct link token information and serves the file directly. It does not suffer from the real direct link limitations regarding cross site requests.


lib/Service/DirectDownloadService.php
```diff
@@ -0,0 +1,68 @@
...
+			if ($nowTs <= $dbExpiration) {
+				$userFolder = $this->root->getUserFolder($dbUserId);
+				$files = $userFolder->getById($dbFileId);
+				if (count($files) > 0 && $files[0] instanceof File) {
+					return $files[0];
+				}
+			}
+		}
+		$req->closeCursor();
+		$qb->resetQueryParts();
```
individual-it, line 64: do those methods also have to be called when returning a file? Because currently they are only called when nothing was found

eneiluj: No they could even be removed. They are useful only if you make other DB requests in the same PHP process.

I always put them so that I never forget them when they are needed :grin:.

In this case I would say we can keep them in case in the future the DirectDownloadController::directDownload() method do something else involving DB requests.

But then you're right, they need to be called before returning the file as well.

individual-it (approved):
Generally works. Some comments that could be solved in a different PR:
1. in the directDownloadPage there is no error shown if the token is invalid, user can still click on the button but then only an empty page appears
2. maybe we actually don't need the directDownloadPage
3. in both cases directDownloadPage and directDownload the file-name does not need to be the original/current filename but its an arbitrary string that will be set as file-name for the downloaded file. Is that a bug or feature

eneiluj:
@individual-it 
1. It was just a way to avoid cookie issues in cross-site contexts. There was no attempt to validate the token indeed :grin:
2. Indeed we don't. DirectDownload page has been removed
3. It's a feature, just like the public share links where it's possible to prepend any string as the last path part to help the browser getting a default name for the downloaded file.

individual-it (approved):

PR #168, opened 2022-07-13
[#43173] use 'typeahead' filter to search for WP

individual-it:
using the `typeahead` filter for the search has several advantages

1. we need only one request and not multiple
2. it also searches for WP ids (fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/43173/)

To make it possible to search also for short ids, the min input is one character

P.S. the global search also now searches for WP ids, but that still has a minimum input of 2 characters

### DIFF
--- a/tests/jest/components/tab/SearchInput.spec.js
+++ b/tests/jest/components/tab/SearchInput.spec.js
@@ -77,6 +77,11 @@ describe('SearchInput.vue tests', () => {
 	describe('work packages multiselect', () => {
 		describe('search input', () => {
 			it('should reset the state if search value length becomes lesser than search char limit', async () => {
+				const axiosSpy = jest.spyOn(axios, 'get')
+					.mockImplementationOnce(() => Promise.resolve({
+						status: 200,
+						data: [],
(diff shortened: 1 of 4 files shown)


individual-it:
CC @Kharonus

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -87,30 +92,27 @@ describe('SearchInput.vue tests', () => {
...
+					search: 'o',
+					expectedCallCount: 1,
 				},
 				{
-					search: 'orga',
+					search: 'or',
 					expectedCallCount: 1,
 				},
-			])('should send search request only if the search text is greater than search threshold', async ({
+			])('should send search request only if the search text is >= than search char limit', async ({
```
kiranparajuli589, line 115:
```suggestion
			])('should send search request only if the search text is greater than or equal to the search char limit', async ({
```
IMO, this will be more readable.

kiranparajuli589 (approved):
lgtm :+1:

PR #171, opened 2022-07-14
[#43199] use host.docker.internal instead of network_mode: "host"

individual-it:
using the host network resulted in port 80 being blocked on the docker host

the issue is that the NextCloud UI (from the users browser, most likely on the docker host) needs to reach OpenProject using the same URL as the NextCloud backend (fron inside the docker service)
this is now achieved by using `host.docker.internal`

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,13 +16,15 @@ Requirements:
   - for v2, make sure to use `docker compose` instead of `docker-compose`
 - OpenProject server instance running in the host machine
 
-  it must be reachable by the hostname `host.docker.internal`.
+  It must be reachable by the hostname `host.docker.internal`, you can do this through this environment variable: `OPENPROJECT_DEV_EXTRA_HOSTS=host.docker.internal`.
 
-  This hostname will resolve to `127.0.0.1` on the docker host and to something like `172.17.0.1` inside of the docker services, so OpenProject needs to listen to those addresses
+  This hostname will resolve to `127.0.0.1` on the docker host and to something like `172.17.0.1` inside of the docker services, so OpenProject needs to listen to those addresses. For that use the `-b` option if you are running OpenProject manually e.g. `bin/rails server -b 0.0.0.0` or when using foreman set `HOST=0.0.0.0` as env. variable.
+
+  The whole OpenProject start command might look something like:
+  - manual start: `OPENPROJECT_DEV_EXTRA_HOSTS=host.docker.internal RAILS_ENV=development bin/rails server -b 0.0.0.0`
+  - using foreman: `HOST=0.0.0.0 OPENPROJECT_DEV_EXTRA_HOSTS=host.docker.internal foreman start -f Procfile.dev`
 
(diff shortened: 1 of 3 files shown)


Kharonus:
Yep, this works as intended. LGTM! 👍🏽

SwikritiT (commented):
https://github.com/nextcloud/integration_openproject/blob/a3b2df6f9dc69287234f835cf11d85efdd3c363e/README.md?plain=1#L46 This can also be removed?

individual-it:
this PR will allow to set `host.docker.internal` as hostname for OpenProject via `OPENPROJECT_DEV_EXTRA_HOSTS` env. variable https://github.com/opf/openproject/pull/10956

SwikritiT:
trying to unlink the package gives me `400` from our side
```
"Server error: `GET http:\/\/host.docker.internal:3000\/api\/v3\/work_packages\/11\/file_links` resulted in a `500 Internal Server Error` response:\n{\"_type\":\"Error\",\"errorIdentifier\":\"urn:openproject-org:api:v3:errors:InternalServerError\",\"message\":\"An internal error  (truncated...)\n"
```

README.md
```diff
@@ -14,9 +14,17 @@ Requirements:
 - Docker Compose
   - for v1, minimum version required is v1.29.0 (our guide is for v1)
   - for v2, make sure to use `docker compose` instead of `docker-compose`
-- OpenProject server instance running in the host machine, you can set up openProject and get it running with the help of this [documentation](https://www.openproject.org/docs/development/development-environment-ubuntu/)
+- OpenProject server instance running in the host machine
 
->**Note:**  While starting the OpenProject server make sure to add environment variable `OPENPROJECT_FEATURE__STORAGES__MODULE__ACTIVE=true` or set `feature_storages_module_active: true` in the `configuration.yml`
```
individual-it, line 19: this seems not to be needed anymore

README.md
```diff
@@ -56,7 +63,7 @@ docker-compose up
 ```
 **Note:** If you've cloned the integration app anywhere other that the default `./../../custom_apps`, provide its path in the `APP_DIR` environment variable
 ```shell
-APP_DIR = <path-to-integration-app> docker-compose up
+APP_DIR=<path-to-integration-app> docker-compose up
 ```
 
 After this, you should be able to access the Nextcloud server at [http://localhost](http://localhost).
```
SwikritiT, line 69:
```suggestion
After this, you should be able to access the Nextcloud server at [http://localhost:8080](http://localhost).
```

SwikritiT (commented):
LGTM

Kharonus:
I have 2 more wishes:
- can we add a docker network to the compose file? It would make integrating this compose file into more complex setups more easy. It can have a simple name like `internal`, `nextcloud`, or even `network`
- can we add an entry called `docker-compose.override.yml` to the `.gitignore`? This would make it more easy to maintain some custom overrides to this setup, like env variable seeding or additional networks.

individual-it:
@Kharonus done, please review again

Kharonus (approved):
LGTM, thanks a lot

PR #173, opened 2022-07-14
[#43201] show different error message for invalid URL

individual-it:
the error message for the case that the URL is invalid should be different


kiranparajuli589 (approved):

individual-it:
this fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/43201

PR #174, opened 2022-07-14
Fix link sanitation, allow target="_blank"

eneiluj:
There was a double sanitation from t() and dompurify.
This removes the t() one and tells dompurify to allow `target="_blank"`

## Related:
fixes: https://community.openproject.org/projects/nextcloud-integration/work_packages/43168/github?query_id=3407


kiranparajuli589:
can we sanitize the text but allow the `target` using the translate fn. if yes, we need not have to sanitize in the TextInput component.

eneiluj:
> can we sanitize the text but allow the target using the translate fn?

There is no such option in the translation function.

https://github.com/nextcloud/server/blob/master/core/src/OC/l10n.js#L113-L151

No option can be passed to dompurify in t() or n().

But it's seems fine to me, we only sanitize once now.

kiranparajuli589 (approved):
lgtm :+1:

PR #176, opened 2022-07-14
calculate contrast for wp type and show border if low

individual-it:
check the contrast between the wp-type color and the background and set `-webkit-text-stroke` if the contrast is too little
1. if the workpackage is not listed yet, get the background color from the whole tab
2. if the workpackage is listed and an hover event happens, recalculate again

part of https://community.openproject.org/projects/nextcloud-integration/work_packages/43206

test-cases:
1. light theme and type color=white/light gray
before:
image
image
now:
image
image
3. dark theme and type color=black
before:
image
now:
image
4. high contrast mode and color=same as hover color of wp
before:
image
now:
image

### DIFF
--- a/src/components/tab/WorkPackage.vue
+++ b/src/components/tab/WorkPackage.vue
@@ -56,7 +56,7 @@ export default {
 	data: () => ({
 		wpTypeTextStroke: 'unset',
 		wpStatusFontColor: '#FFFFFF',
-		wpStatusBorder: '0px',
+		wpStatusBorder: '0',
 	}),
 	created() {
 		this.setWPTypeTextStroke()
@@ -100,7 +100,7 @@ export default {
 					this.wpStatusBorder = 'solid 1px #000000'
 					return
 				}
-				this.wpStatusBorder = '0px'
+				this.wpStatusBorder = '0'
 			} catch (e) {
 				// something went  wrong, leave the values as they are
 			}
--- a/src/components/tab/WorkPackage.vue
+++ b/src/components/tab/WorkPackage.vue
@@ -2,8 +2,9 @@
 	<div class="workpackage" @mouseover="setWPTypeTextStroke" @mouseleave="setWPTypeTextStroke">
 		<div class="row">
 			<div class="row__status"
-				:style="{'background-color':workpackage.statusCol}">
-				<div class="row__status__title">
+				:style="{'background-color': getWPStatusColor()}">
+				<div class="row__status__title"
+					:style="{'color': wpStatusFontColor }">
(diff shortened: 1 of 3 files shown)


wielinde:
Probably not relevant for this PR as this PR deals with OpenProject types. Here I found an issue when having a black status 
image

wielinde:
@individual-it I have played around with various work package type colors and light and dark themes in Nextcloud. And it works pretty well. As always, I didn't check the code.

individual-it:
OP sets the font color to white if the status color is dark
image
image

I think we can do some CSS for that, I will create an extra PR

individual-it:
the pure CSS solution did not produce the results I wanted, so I used the same contrast calculation also for the status font color
image
image
image

we also could draw a box around the status if there is no contrast between the status and the background of it, OP does not do it
image

individual-it:
boarders are now shown if there is not enough contrast between statuscolor and background
examples
image
image
image

no border with enough contrast
image
image


Please test this well, I failed so far to write meaningful automated tests for this feature

kiranparajuli589 (commented):
when I use a status color as `No Color`, the orange color is on the NC side status. is this expected?

individual-it:
@kiranparajuli589 yes, see `Bug #2` in https://community.openproject.org/projects/nextcloud-integration/work_packages/43206
it is the same default color as used in OP is no color is set

src/components/tab/WorkPackage.vue
```diff
@@ -50,6 +53,171 @@ export default {
 			required: true,
 		},
 	},
+	data: () => ({
+		wpTypeTextStroke: 'unset',
+		wpStatusFontColor: '#FFFFFF',
+		wpStatusBorder: '0px',
```
kiranparajuli589, line 59:
```suggestion
		wpStatusBorder: '0',
```

src/components/tab/WorkPackage.vue
```diff
@@ -50,6 +53,171 @@ export default {
...
+				if (contrast <= 2) {
+					contrast = this.contrastRatio(this.getWPStatusColor(), '#000000')
+					if (contrast <= 2) {
+						this.wpStatusBorder = 'solid 1px #FFFFFF'
+						return
+					}
+					this.wpStatusBorder = 'solid 1px #000000'
+					return
+				}
+				this.wpStatusBorder = '0px'
```
kiranparajuli589, line 103:
```suggestion
				this.wpStatusBorder = '0'
```

kiranparajuli589 (approved):

PR #177, opened 2022-07-18
[#43211] reset OP host validity check

individual-it:
when the host was checked if it served an open project instance but the input was changed the data of the check also needs to be invalidated

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/43211


SwikritiT (approved):
LGTM

PR #178, opened 2022-07-18
[#43160] reset all app values

individual-it:
show an reset button on the top right corner, when clicked show a warning message and when confirmed clear all app settings and connections of every user

image
image
image

currently on top of #180 

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/4316


src/components/AdminSettings.vue
```diff
@@ -369,6 +379,31 @@ export default {
...
+				},
+				true
+			)
+		},
+		async resetAllAppValues() {
+			this.state.client_id = null
+			this.state.client_secret = null
+			this.state.oauth_instance_url = null
+			await this.saveOPOptions()
+			window.location.reload()
```
kiranparajuli589, line 405: if the save has not succeeded, maybe reloading is not necessary.

individual-it: I would still do a reload, that gives the user the real current status, whatever was reset or not

SwikritiT (commented):
LGTM

PR #179, opened 2022-07-19
[#43302] show last modifier in files API

individual-it:
use the activity app to get the latest modifier.
Only content changes of a file will be regarded as modifications, move, rename etc. not.
fields that are returned: `modifier_id` and `modifier_name`
In case the modifier could not be found (e.g. activity app is not installed, last modifier is deleted, public modified last, ...) `null` is returned

for details see API tests

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/43302

### DIFF
--- a/lib/Controller/FilesController.php
+++ b/lib/Controller/FilesController.php
@@ -11,20 +11,24 @@
 
 namespace OCA\OpenProject\Controller;
 
-use OC\User\User;
+use OCA\Activity\Data;
 use OCA\Activity\GroupHelper;
 use OCA\Activity\UserSettings;
 use OCA\Files_Trashbin\Trash\ITrashManager;
 use OCP\Activity\IManager;
 use OCP\Files\Config\IMountProviderCollection;
 use OCP\Files\IRootFolder;
+use OCP\IConfig;
 use OCP\IDBConnection;
+use OCP\IL10N;
+use OCP\ILogger;
 use OCP\IRequest;
 use OCP\AppFramework\OCSController;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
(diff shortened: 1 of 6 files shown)


lib/Controller/FilesController.php
```diff
@@ -144,4 +210,62 @@ private function compileFileInfo($fileId) {
...
+				}
+				if (str_starts_with($activityDetails->getSubject(), 'changed')) {
+					return $this->userManager->get($activity['user']);
+				}
+			}
+		}
+		if ($activities['has_more'] === true) {
+			$lastGiven = (int)$activities['headers']['X-Activity-Last-Given'];
+			if (($lastGiven < $since || $since === 0) && $lastGiven != $since) {
+				return $this->getLastModifier($ownerId, $fileId, $lastGiven);
```
Kharonus, line 266: not sure, why this recursion is needed, but I do not know the data structure. Is this about pagination in the activities data?

Kharonus (commented):
Looks quite fine for me. I will start to use this branch to implement our side of this issue.

lib/Controller/FilesController.php
individual-it: only 10 items are fetched from the DB, see https://github.com/nextcloud/integration_openproject/pull/179/files#diff-776f8feca78c8322be8a9c5a0a5c6dcb93dfdd2776045ee36faa1a4c172d645fR244
I think we could go up to 200 or so. In any case the recursion is needed because the fetched items might all not be a change, so we want to try again till we find some relevant activity

kiranparajuli589 (commented):
lgtm :+1:

kiranparajuli589 (approved):

PR #180, opened 2022-07-19
[#43303] only clean userinfo when client id or secret changed

individual-it:
if only the oauth URL is changed we don't want to delete the user oauth tokens and make everyone to login again, instead if the server is moved the users should not be affected

how to test:
1. setup NC-OP connection
2. connect a user
3. change the OP URL in NC (e.g. from `http://localhost:3000` to `http://127.0.0.1:3000`)
4. check if the user is still connected

the only  "issue" with this approach is, that when the admin clicks on "Replace OpenProject OAuth values" the app shows a message saying that every user will have to login again, BUT if the admin enters again exactly the same id and secret nothing will happen

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -138,7 +138,15 @@ class ConfigController extends Controller {
 	 * @return DataResponse
 	 */
 	public function setAdminConfig(array $values): DataResponse {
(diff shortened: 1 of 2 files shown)


kiranparajuli589 (commented):
works as expected for me :+1:

SwikritiT (approved):
Working for me as well

PR #181, opened 2022-07-20
unify highest level of test description

individual-it:
make the higest level of the JS test-description look the same and fix the one that is wrong


kiranparajuli589 (commented):
snapshots also need to be updated after updating the test description

individual-it:
@kiranparajuli589 done

kiranparajuli589 (approved):

PR #184, opened 2022-07-26
run tests with NC22

individual-it:
the app claims to have NC 22 as min version, so lets test it

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,11 @@ jobs:
     name: CI
     strategy:
       matrix:
-        nextcloudVersion: [ v23.0.4, v24.0.0, master ]
+        nextcloudVersion: [ v22.2.9, v23.0.4, v24.0.0, master ]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -11,9 +11,11 @@ jobs:
     name: CI
     strategy:
       matrix:
-        nextcloudVersion: [ v23.0.4, v24.0.0, master ]
+        nextcloudVersion: [ v22.2.9, v23.0.4, v24.0.0, master ]
```
eneiluj, line 14:
```suggestion
        nextcloudVersion: [ stable22, stable23, stable24, master ]
```
It will save you some effort to adjust to latest minor releases.

eneiluj (changes requested):

.github/workflows/ci.yml
individual-it: thanks, adjusted

kiranparajuli589 (approved):

PR #188, opened 2022-07-28
New public endpoint returning the user id and the authorization header

eneiluj:
Observed behaviour:
* when using basic auth with invalid credentials, the request does not reach the app controller and we get a 401 response with `{"message":""}` in the body
* successful basic auth: `{"user_id":"USER_ID","authorization_header":"Basic anVsaWVuOk9yYW4nZV9jbG91sA=="}`
* passing an invalid bearer token with `Authorization: Bearer INVALID_TOKEN`: we get `{"user_id":"","authorization_header":"Bearer INVALID_TOKEN"}`
* passing a valid token, we get `{"user_id":"USER_ID","authorization_header":"Bearer VALID_TOKEN"}`

https://community.openproject.org/wp/43473

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -15,6 +15,7 @@ return [
 		['name' => 'config#setConfig', 'url' => '/config', 'verb' => 'PUT'],
 		['name' => 'config#setAdminConfig', 'url' => '/admin-config', 'verb' => 'PUT'],
 		['name' => 'config#autoOauthCreation', 'url' => '/nc-oauth', 'verb' => 'POST'],
+		['name' => 'config#checkConfig', 'url' => '/check-config', 'verb' => 'GET'],
 
 		['name' => 'directDownload#directDownload', 'url' => '/direct/{token}/{fileName}', 'verb' => 'GET'],
 
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -334,4 +334,18 @@ class ConfigController extends Controller {
 			$this->config->deleteAppValue(Application::APP_ID, 'nc_oauth_client_id');
 		}
 	}
+
+	/**
+	 * @NoCSRFRequired
+	 * @NoAdminRequired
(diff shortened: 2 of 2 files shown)


individual-it:
@Kharonus is this API what you need?

wielinde:
@eneiluj @Kharonus @individual-it Thank you for this endpoint. I believe there is one important piece missing: It actually does not tell me that the application is installed when it strips the Authorization header as it would not be passed on to the app controller. So from the OpenProject side I cannot tell if the problem is a missing installation of the app or a bad configuration of Apache. Then, how could I help the user to debug the situation? In most cases the application is simply missing and I don't want to bother them with Apache configuration tips.

Kharonus:
Wait @wielinde , isn't there a 404 if the app is not installed? because the endpoint is not available?

What is kind of unclear for me: What response do we expect, when the bearer token is stripped due to missing config?

eneiluj:
@wielinde If the app is not installed, you will receive a 302 response. 2 cases:
* you are not authenticated: you'll get this HTTP response header: `Location: https://your.nextcloud.org/index.php/login`
* you are authenticated (any kind of auth): you'll get this HTTP response header: `Location: https://your.nextcloud.org/index.php/apps/dashboard/`

wielinde:
@eneiluj @individual-it OK, understood. The only thing I didn't manage to achieve was having index.php paths everywhere but Authorization header being passed (I think that is the scenario that Artur is having in his private production instance). So I cannot test that now. For the time being I will simply enable and disable mod_rewrite.

individual-it:
@wielinde I've managed to have the  auth headers working but having `index.php` everywhere by enabling mod_rewrite but resetting the `.htaccess` file to the default that in in the NC repo
The check-config API behaves as expected also in that case

individual-it (approved):

wielinde:
@individual-it I am not done with the implementation on the OP side that is using it. But it looks to me that everything I need is in place. So please merge.

PR #193, opened 2022-08-05
[#43473] add index.php always to the redirect URL

individual-it:
NC needs always to send index.php  in the redirect url, even if it runs currently without, because OP always expects the URLs to include index.php

this depends on https://github.com/opf/openproject/pull/11086

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -236,15 +236,12 @@ class ConfigController extends Controller {
 		}
 
 		if ($clientID && $validClientSecret && $validCodeVerifier && $configState !== '' && $configState === $state) {
-			$redirectUri = $this->urlGenerator->getAbsoluteURL(
-				'/index.php/apps/' . Application::APP_ID . '/oauth-redirect'
-			);
 			$openprojectUrl = $this->config->getAppValue(Application::APP_ID, 'oauth_instance_url');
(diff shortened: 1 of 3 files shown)


lib/Controller/ConfigController.php
```diff
@@ -236,15 +236,15 @@ public function oauthRedirect(string $code = '', string $state = ''): RedirectRe
 		}
 
 		if ($clientID && $validClientSecret && $validCodeVerifier && $configState !== '' && $configState === $state) {
-			$redirect_uri = $this->urlGenerator->linkToRouteAbsolute(
-				Application::APP_ID . '.config.oauthRedirect'
+			$redirectUri = $this->urlGenerator->getAbsoluteURL(
+				'/index.php/apps/' . Application::APP_ID . '/oauth-redirect'
```
wielinde, line 240: @individual-it Minor remark: Without knowing too much about how you organize code in PHP.... it looks a bit suspicious that you have two places here in the code doing exactly the same. In `lib/Service/OpenProjectAPIService.php` you generate exactly the same URL again.

individual-it: @wielinde you are right, I've put it into a separate function
I also suspect that we don't need to have some of the code static, but I don't want to refactor it for 2.0.0, I will create an other WP for it

SwikritiT:
It's working for me

SwikritiT (approved):

PR #194, opened 2022-08-07
Update npm and node engines versions

nextcloud-command:
Automated update of the npm and node engines versions


kiranparajuli589 (commented):
changes lgtm :+1: @individual-it @eneiluj should we merge this?

individual-it:
that will simply require node 16 for building the extension, or will anything break?

kiranparajuli589:
> that will simply require node 16 for building the extension, or will anything break?

yes, the engine requirement should be fulfilled for the building purpose. but not sure if anything breaks. One good thing is that current tests are happy at least.

PR #198, opened 2022-08-10
use name of file in context of requesting user

individual-it:
for debug

PR #199, opened 2022-08-11
harden getting mounts and fall back to requester context

individual-it:
alternative for #198

the code now loops though all mounts and tries to identify the mount of the owner, because we want to get the file name in the context of the owner
if for some reason that is not possible it falls back to return the file name in the context of the requester


wielinde:
@individual-it Looks good on our edge environment. For us it is good to merge.

wielinde (approved):
I've tested it on our Edge instance and it seems to solve the problem. Thank you.

PR #203, opened 2022-08-17
[OP#43666] Do not show the search bar if personal settings is not complete

SwikritiT:
OP#43666 https://community.openproject.org/projects/nextcloud-integration/work_packages/43666/activity


individual-it (changes requested):
I think we need to make it even more general, it still shows up if the connection is setup correctly but OP is not reachable
image

SwikritiT:
> I think we need to make it even more general, it still shows up if the connection is setup correctly but OP is not reachable image

Done in latest commit can you check again!

kiranparajuli589 (commented):
lgtm :+1:

individual-it (commented):
LGTM, let's merge that after 2.0.2 is out

individual-it (approved):

PR #205, opened 2022-08-17
[OP#42323] aggregate notifications in the dashboard by WP and show more info

individual-it:
The dashboard shows the notifications aggregated by work package, with some extra information and all notifications of a work package can be marked as read at once

image


src/views/Dashboard.vue
```diff
@@ -146,11 +157,53 @@ export default {
 			this.fetchNotifications()
 			this.loop = setInterval(() => this.fetchNotifications(), 60000)
 		},
+		getNotificationsUrl() {
+			return generateUrl('/apps/integration_openproject/notifications')
+		},
```
kiranparajuli589, line 162: this can be used as a computed property so that we do not have to make a call to get this value.

tests/jest/views/Dashboard.spec.js
```diff
@@ -0,0 +1,194 @@
...
+			.mockImplementationOnce(() => Promise.resolve({ data: notificationsResponse }))
+		await wrapper.vm.launchLoop()
+		expect(axiosSpy).toBeCalledWith(
+			'http://localhost/apps/integration_openproject/url',
+		)
+		expect(axiosSpy).toBeCalledWith(
+			'http://localhost/apps/integration_openproject/notifications',
+		)
+		expect(wrapper.vm.state).toBe(STATE.OK)
+		expect(wrapper.vm.items).toStrictEqual([
```
kiranparajuli589, line 64: maybe saving as the snapshot is better. that way long expected array is not necessary. also easy to maintain.
```suggestion
		expect(wrapper.vm.items).toMatchSnapshot()
```

tests/jest/views/Dashboard.spec.js
```diff
@@ -0,0 +1,194 @@
...
+					},
+					propsData: {
+						title: 'dashboard',
+					},
+				})
+			axiosSpyGet = jest.spyOn(axios, 'get')
+				.mockImplementation(() => Promise.resolve({
+					data: [
+						{
+							_type: 'Notification',
```
kiranparajuli589, line 120: can we get this item from the fixture you've created?

src/views/Dashboard.vue
individual-it: it actually is only needed in one place, so moved it to a local variable

tests/jest/views/Dashboard.spec.js
individual-it: done! I was not aware that we can use snapshot on a arbitrary object, but that totally makes sense

tests/jest/views/Dashboard.spec.js
individual-it: done, good point!

kiranparajuli589 (approved):
lgtm :+1:

PR #209, opened 2022-08-19
Updated description for the OpenProject app

psatyal:
https://community.openproject.org/projects/nextcloud-integration/work_packages/43715/activity

Updated the copy of the description to make it clearer, easier to read and more to-the-point

Signed-off-by: Parimal Satyal <88370597+psatyal@users.noreply.github.com> (OpenProject)

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -7,7 +7,7 @@
 	
 * Link files and folders with work packages in OpenProject
 * Find all work packages linked to a file or folder
-* Keep an eye on updates to linked work packages via the dashboard
+* Keep an eye on updates to your work packages via the dashboard
 * Find relevant work packages using Nextcloud's search bar
 
 On the OpenProject end, users are able to:
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -4,7 +4,7 @@
 	<name>OpenProject integration</name>
 	<summary>Integration of OpenProject project management tool</summary>
 	<description><![CDATA[This application enables seamless integration with open source project management and collaboration software OpenProject. On the Nextcloud end, it allows users to:
-
+	
 * Link files and folders with work packages in OpenProject
 * Find all work packages linked to a file or folder
 * Keep an eye on updates to linked work packages via the dashboard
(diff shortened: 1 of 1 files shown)


appinfo/info.xml
```diff
@@ -3,7 +3,21 @@
...
 	<name>OpenProject integration</name>
 	<summary>Integration of OpenProject project management tool</summary>
-	<description><![CDATA[This application deeply integrates with the open source project management software OpenProject. It allows linking files and folders in Nextcloud with work packages in OpenProject. In OpenProject, you can see all files related to a work package, and in Nextcloud, you can find all work packages linked to a file. Your project teams gain productivity as they clearly know where to find and store the files that they need for finishing a work package. Further, it extends the Nextcloud dashboard with a widget for personal notifications from OpenProject. The app can also extend the search function in Nextcloud to find work packages in OpenProject.]]></description>
+	<description><![CDATA[This application enables seamless integration with open source project management and collaboration software OpenProject. On the Nextcloud end, it allows users to:
+	
+* Link files and folders with work packages in OpenProject
+* Find all work packages linked to a file or folder
+* Keep an eye on updates to linked work packages via the dashboard
```
individual-it, line 10: This is not completely correct, the dashboard shows all OP notifications, not only updates of linked WP

appinfo/info.xml
```diff
@@ -3,7 +3,21 @@
...
+* Find relevant work packages using Nextcloud's search bar
+
+On the OpenProject end, users are able to:
+
+* View all Nextcloud files linked to a work package
+* Download linked files or open them in Nextcloud to edit them
+* Unlink linked files
+
+For more information on how to set up and use the OpenProject application, please refer to [integration setup guide](https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/) for administrators and [the user guide](https://www.openproject.org/docs/user-guide/nextcloud-integration/).
```
individual-it, line 19: I'm not sure if the links will be rendered correctly in the app-store. @eneiluj do you know.
But maybe it's enough that the links are anyway shown in the resources area
image

individual-it (changes requested):

appinfo/info.xml
psatyal: Important distinction, thanks. Updated the text.

psatyal (commented):
Made changes based on @individual-it's comments.

individual-it (approved):

individual-it:
CI fails because code-coverage does not work correctly when the PR comes from an other repo. But in this case that does not matter at all

PR #215, opened 2022-08-22
[OP#43667] fix redirect urls for the new dedicated settings section

kiranparajuli589:
Description: Fix the redirect urls for the new dedicated personal settings section

Related: https://community.openproject.org/projects/nextcloud-integration/work_packages/43667

---


individual-it (approved):

individual-it (changes requested):
still does not work

individual-it (changes requested):
you also need to change the code that makes sure the user is returned to the correct starting page https://github.com/nextcloud/integration_openproject/blob/master/lib/Controller/ConfigController.php#L207

kiranparajuli589:
> you also need to change the code that makes sure the user is returned to the correct starting page https://github.com/nextcloud/integration_openproject/blob/master/lib/Controller/ConfigController.php#L207

fixed

individual-it (approved):

PR #217, opened 2022-08-23
[#43824] warn admin when using local OP instance

individual-it:
by default NC does not allow to connect to other "local" servers, to allow that the `allow_local_remote_servers` setting needs to be enabled.
Even that should not happen much on production systems, admins who just want to try the integration might just want to run everything on "localhost" or similar.

with this code changes the app detects that the request was refused because it was send to a local server (by catching the correct Exception) and shows a message that tells the admin what happened and where to find help

image


individual-it:
proposed documentation change in https://github.com/opf/openproject/pull/11187

kiranparajuli589 (commented):
lgtm :+1:

individual-it:
distinguish more errors according to to exceptions see #229
point the user to the logs

src/components/admin/TextInput.vue
```diff
@@ -31,10 +31,16 @@
...
-				{{ errorMessage }}
+			<div v-if="errorMessage">
+				<div class="text-input-error-message"
+					v-html="sanitizedErrorMessage" /> <!-- eslint-disable-line vue/no-v-html -->
+				<Popover v-if="errorMessageDetails">
+					<template #trigger>
+						<a class="link" href="#">{{ t("integration_openproject", "Details") }}</a>
+					</template>
+					<div v-html="sanitizedErrorMessageDetails" /> <!-- eslint-disable-line vue/no-v-html -->
+				</Popover>
```
kiranparajuli589, line 42: for the popover content, imo a little bit more padding is needed. also, the details button in the second line is a bit odd. This is just my opinion. If we can use some dots icon button or hovering the error message to show the tooltip would be good.

individual-it: added some padding and moved the "Details" link to the right of the Error message (using float, is that the correct way to do it?)
image

individual-it:
hovering would not work, because in some details we want to have a link to the docs, except if we place the link outside of the popup

kiranparajuli589 (commented):
LGTM :+1: checked with NC stable25, works as expected in my machine.

kiranparajuli589 (approved):

PR #223, opened 2022-08-25
[OP#43669] allow the administrator to define defaults for user settings

kiranparajuli589:
### Description
- Added default values form in the Admin Section
   
    Screenshot from 2022-08-25 14-54-12

https://user-images.githubusercontent.com/41103328/191679680-cda4f734-7b40-4e17-ae82-e89a4816509e.mp4

- Update the `/admin-config` route to update the following admin configs too.
  - default_enable_unified_search
  - default_enable_notifications
  - default_enable_navigation

- Make use of the default config if the user has not set any personal values.

### Related
- Part of https://community.openproject.org/projects/nextcloud-integration/work_packages/43669


src/components/AdminSettings.vue
```diff
@@ -286,29 +304,37 @@ export default {
...
-			}
-			if (this.state.nc_oauth_client) {
-				this.formMode.ncOauth = F_MODES.VIEW
-				this.isFormCompleted.ncOauth = true
-			}
-		}
+		this.init()
 	},
 	methods: {
+		init() {
```
individual-it, line 317: why moving that code to an own function?

kiranparajuli589: imo, it's better to call an fn from the hooks.

SwikritiT:
I have updated the description with the current implementation video

src/components/AdminSettings.vue
```diff
@@ -398,9 +426,21 @@ export default {
...
+			this.isFormCompleted.opOauth = false
+			this.formMode.server = F_MODES.EDIT
+			this.isFormCompleted.server = false
+
 			this.state.client_id = null
 			this.state.client_secret = null
 			this.state.oauth_instance_url = null
+			this.state.default_enable_navigation = null
+			this.state.default_enable_notifications = null
+			this.state.default_enable_unified_search = null
```
eneiluj, line 442: We could set it back to false to keep it as a boolean.

src/components/AdminSettings.vue
```diff
@@ -441,7 +484,6 @@ export default {
 				showSuccess(t('integration_openproject', 'OpenProject admin options saved'))
 				return true
 			} catch (error) {
-				console.debug(error)
```
eneiluj, line 444: As you want but I think it's good to have a console.error there.

eneiluj (changes requested):

individual-it (changes requested):
this works nicely to show the correct settings in the admin and personal settings page
BUT: where we use the settings for real to trigger the actions only the usersettings are read, not the defaults
e.g https://github.com/nextcloud/integration_openproject/blob/caa0b817395c067dd1e53e53016ff6df3f3f779c/lib/Service/OpenProjectAPIService.php#L137
https://github.com/nextcloud/integration_openproject/blob/86987d0632d38fef9061e283d03f05598b14ab04/lib/Search/OpenProjectSearchProvider.php#L130
https://github.com/nextcloud/integration_openproject/blob/bccb09abd9326c1988f253862a28621a33810465/lib/AppInfo/Application.php#L86

SwikritiT:
> this works nicely to show the correct settings in the admin and personal settings page BUT: where we use the settings for real to trigger the actions only the usersettings are read, not the defaults e.g
> 
> https://github.com/nextcloud/integration_openproject/blob/caa0b817395c067dd1e53e53016ff6df3f3f779c/lib/Service/OpenProjectAPIService.php#L137
> 
> 
> https://github.com/nextcloud/integration_openproject/blob/86987d0632d38fef9061e283d03f05598b14ab04/lib/Search/OpenProjectSearchProvider.php#L130
> 
> 
> https://github.com/nextcloud/integration_openproject/blob/bccb09abd9326c1988f253862a28621a33810465/lib/AppInfo/Application.php#L86

Can you try again? It should work now.

individual-it (approved):
looks good, please adjust the tests

SwikritiT:
@eneiluj I've addressed your reviews can you take a look again?

PR #224, opened 2022-08-25
mirror OpenProject notifications to Nextcloud

individual-it:

All current notifications of OpenProject should be mirrored to Nextcloud.
The view is similar to the notification center in OpenProject where the notifications are aggregated by work-package.
Every time the notifications are fetched again from OpenProject, the Nextcloud notifications are dismissed and recreated according to what is the current status in OpenProject.
The Nextcloud dismiss button has basically no functionality, but through the "Mart as Read" button the user can mark all notifications of that work-package as read and they will not be displayed neither in Nextcloud nor in OpenProject

image

### DIFF
--- a/lib/BackgroundJob/CheckNotifications.php
+++ b/lib/BackgroundJob/CheckNotifications.php
@@ -47,7 +47,7 @@ class CheckNotifications extends TimedJob {
 								LoggerInterface $logger) {
 		parent::__construct($time);
 		// Every 15 minutes
-		$this->setInterval(60 * 15);
+		$this->setInterval(10);
 
 		$this->openprojectAPIService = $openprojectAPIService;
 		$this->logger = $logger;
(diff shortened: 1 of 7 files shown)


lib/BackgroundJob/CheckNotifications.php
```diff
@@ -47,7 +47,7 @@ public function __construct(ITimeFactory $time,
 								LoggerInterface $logger) {
 		parent::__construct($time);
 		// Every 15 minutes
-		$this->setInterval(60 * 15);
+		$this->setInterval(10);
```
individual-it, line 50: need to be reverted

wielinde:
@individual-it I am sorry to bother you. From the title of the PR I don't get what this is about. Maybe I don't need to know it?

individual-it:
@wielinde the description would not helped ether ;-)
This is still WIP, so I wanted to write a better text when finished, but I guess I can do it straight away.

individual-it:
this needs #205 so the PR is towards `improveNotifications`
I think as soon #205 is merged we can create a new PR towards master. But still this PR here can be already reviewed

individual-it:
closing in favor of #256

PR #228, opened 2022-09-07
Fix trashManager not found

eneiluj:
Only use the trash manager if the files_trashbin app is enabled for the user.
Catch any exception when getting the trashManager, just in case.

https://community.openproject.org/wp/43936

### DIFF
--- a/lib/Controller/FilesController.php
+++ b/lib/Controller/FilesController.php
@@ -11,11 +11,12 @@
 
 namespace OCA\OpenProject\Controller;
 
+use Exception;
 use OCA\Activity\Data;
 use OCA\Activity\GroupHelperDisabled;
 use OCA\Activity\UserSettings;
-use OCA\Files_Trashbin\Trash\ITrashManager;
 use OCP\Activity\IManager;
+use OCP\App\IAppManager;
 use OCP\Files\Config\ICachedMountFileInfo;
 use OCP\Files\Config\IMountProviderCollection;
 use OCP\Files\FileInfo;
@@ -32,6 +33,7 @@ use OCP\IUser;
 use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\RichObjectStrings\IValidator;
+use Throwable;
 
 class FilesController extends OCSController {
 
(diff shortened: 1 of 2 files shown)


eneiluj:
The phpstan error can be ignored as it is the way to go :grin:.

lib/Controller/FilesController.php
```diff
@@ -158,16 +160,24 @@ public function getFilesInfo(?array $fileIds): DataResponse {
...
 			$file = $files[0];
 			$trashed = false;
-		} else {
-			$file = $this->trashManager->getTrashNodeById(
-				$this->user, $fileId
-			);
-			$trashed = true;
+		} elseif ($this->appManager->isEnabledForUser('files_trashbin')) {
+			try {
+				$trashManager = Server::get(\OCA\Files_Trashbin\Trash\ITrashManager::class);
```
individual-it, line 173: this only works for NC 25, see https://github.com/nextcloud/server/commit/f945c0cbc6aef461fabf17cea42440ad9b6fea09#diff-e3c727c87edec3701539efc3b38c11374baf33362c0a4038f004f4b7ccbe71b7R39

That is why phpstan complains with older versions.

individual-it (changes requested):

lib/Controller/FilesController.php
eneiluj: Alright, fixed. Could you adjust the tests?

individual-it:
strangely the unit tests work fine when run separately but when ran all at once `getTrashNodeById`  throw strange exceptions  `"Backends provided no user object for testUser"`
need to investigate more

individual-it:
tests are fixed and some new ones added.
there are two cases
1. a file got deleted, then the trashbin app got disabled
    in that case the file actually is in the trashbin and has a mount, but cannot be accessed, the return code is 403
2. a file got deleted after the trashbin app got disabled
    in that case the file is really gone, the return code is 404

individual-it (commented):
I think we can merge that
@eneiluj are you happy with my refactoring?
@wielinde should that be backported to 2.0 ?

wielinde:
@individual-it Yes, I think this PR is a typical for a patch level release

eneiluj:
@individual-it Yep, all good.

individual-it (approved):

individual-it:
backport in  #232

PR #230, opened 2022-09-15
Use auto-release branch for automatic releases, drop the 'release' one

eneiluj:
Once this is merged, we can delete the "release" branch.

:warning: This commit needs to be in the branch which is then pushed in "auto-release", otherwise the action won't be triggered. Be careful because once merged, even if it's in master, it's not in any other branch.

If needed we can make other PRs bringing this change in other branches.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [release]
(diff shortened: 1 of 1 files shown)


individual-it (approved):

individual-it:
backport in #231

PR #234, opened 2022-09-21
Simplify css and fix icon style for NC 25

eneiluj:
Icon style was wrong for NC 25.

Here is a way to avoid the conditional class for the dashboard/files-app icon.
It works in NC 24 and 25 with darkmode enabled or disabled.

`.option-open-project > .icon-openproject,` fixes the right click menu style.


SwikritiT:
LGTM

individual-it (approved):

PR #238, opened 2022-09-23
run CI with stable25 branch of NC

individual-it:
the master branch on NC got bumped to version 26, the activity app does not support that. For now we should run the tests with the stable25 branch


SwikritiT (approved):
LGTM

individual-it:
backport in #239

PR #241, opened 2022-09-27
[OP#44145] do not translate oauth technical terms

individual-it:
OAuth ID and secret are technical terms that should be the same in all languages. Do not translate them.

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/44145


kiranparajuli589 (approved):
lgtm :+1:

PR #245, opened 2022-09-29
build 2.0 using NC 24

individual-it:
Nextcloud 24 is the current stable, so I suggest to use that to build 2.0 of our integration


wielinde (approved):
Looks like a solution for now. However, we need to revise it as soon as we can get hold of @eneiluj .

PR #249, opened 2022-10-13
[#43927] Replace the word "notifications" in dashboard header with a "bell" icon

kiranparajuli589:
## Description
Add a bell icon in the dashboard card heading

## Related
Fixes: https://community.openproject.org/projects/nextcloud-integration/work_packages/43927

## Snapshots
Screenshot from 2022-10-20 16-38-03
Screenshot from 2022-10-20 16-38-20
Screenshot from 2022-10-20 16-38-36
Screenshot from 2022-10-20 16-38-50

## Tested with NC
- [x] master
- [x] stable25
- [x] stable24
- [x] stable23
- [x] stable22


lib/Dashboard/OpenProjectWidget.php
```diff
@@ -84,7 +84,7 @@ public function getId(): string {
 	 * @inheritDoc
 	 */
 	public function getTitle(): string {
-		return $this->l10n->t('OpenProject notifications');
+		return $this->l10n->t('OpenProject');
```
individual-it, line 87: what about doing something like this:
`return $this->l10n->t('OpenProject') . ' 🔔';`

the bell is an UTF character: https://www.utf8icons.com/character/128276/bell
For me it looks like that:
image

css/dashboard.css
```diff
@@ -1,18 +1,58 @@
...
+.panel--content > div {
+	height: 100%;
+}
+
+.panel--content .empty-content {
+	margin: 0 !important;
+	height: 100%;
+	align-items: start;
+	justify-content: center;
+}
```
kiranparajuli589, line 34: the panel was rendering with big height. there is a top margin used in the empty-content class (maybe from NC vue not from the app code). this CSS should fix that and render panel with normal height in every NC versions.

individual-it (commented):
I like the solution
@wielinde @marcalcobe what do you think?

individual-it (approved):

PR #252, opened 2022-10-13
user master branch of NC for release

individual-it:
in case there will be an other 2.0 release, we descided to build again with NC master branch
for that its also required to increase the max-version on NC


wielinde (approved):
@individual-it Looks good. I saw that the other branches (release/2.1 and master) are already up to date.

individual-it:
@wielinde yes, because we never changed our master branch only release/2.0

PR #253, opened 2022-10-13
React to notification dismiss

julien-nc:
I did some digging in the notification implementation in the core and the notifications apps and, good news, we can know when one of our notifications is dismissed.

The IDismissableNotifier interface adds the `dismissNotification` method to react to notification dismiss.

@individual-it Is this enough to move forward on your side?


individual-it:
wow, that is really great and easy to use, see d56b96f58be7bcceaff01dba835d647659a496fe
I will close this PR in favor of #256
Thank you!

PR #256, opened 2022-10-17
[OP#42323] mirror OP notifications to NC notifications

individual-it:

All current notifications of OpenProject should be mirrored to Nextcloud.
The view is similar to the notification center in OpenProject where the notifications are aggregated by work-package.

- [x] ToDo: don't use the blue "mark as read button" but react on the NC dismiss button. See #253

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,11 @@ jobs:
     name: CI
     strategy:
       matrix:
-        nextcloudVersion: [ stable23, stable24, stable25, master ]
+        nextcloudVersion: [ stable22, stable23, stable24, stable25, master ]
         phpVersion: [ 7.4, 8.0, 8.1 ]
         exclude:
+          - nextcloudVersion: stable22
+            phpVersion: 8.1
           - nextcloudVersion: stable23
             phpVersion: 8.1
           - nextcloudVersion: stable24
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,11 +11,9 @@ jobs:
     name: CI
     strategy:
       matrix:
-        nextcloudVersion: [ stable22, stable23, stable24, stable25, master ]
+        nextcloudVersion: [ stable23, stable24, stable25, master ]
         phpVersion: [ 7.4, 8.0, 8.1 ]
         exclude:
-          - nextcloudVersion: stable22
-            phpVersion: 8.1
           - nextcloudVersion: stable23
             phpVersion: 8.1
           - nextcloudVersion: stable24
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,7 @@ jobs:
           git clone --depth 1 https://github.com/nextcloud/server.git -b ${{ matrix.nextcloudVersion }}
           cd server && git submodule update --init
           ./occ maintenance:install --admin-pass=admin
+          git clone --depth 1 https://github.com/nextcloud/notifications.git -b ${{ matrix.nextcloudVersion }} apps/notifications
 
       - name: PHP stan
         run: make phpstan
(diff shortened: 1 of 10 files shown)


individual-it:
@julien-nc your idea about how to react on the dismiss button works great, but now I have a new problem.
Every time I check for new notifications I want to get rid of all old ones, so I call `markProcessed()` in https://github.com/nextcloud/integration_openproject/pull/256/files#diff-acd9bb903dbec989fbe06b290cfcaed634a3f67d497845d4d6946fa1b6449c93R173
That itself calls the function to mark the notifications as read in OP.
I cannot find a way of distinguish between the user clicking the dismiss button and a call from `markProcessed`

individual-it:
using now `OCA\Notifications\Handler::deleteIds()` to delete existing notifications before recreating them. This function does not trigger the `dismissNotification()` hook so its ideal for us, but it was first introduced in NC 23 (https://github.com/nextcloud/notifications/pull/1156), so using that function would mean dropping support for NC 22

individual-it:
`deleteIds()` was not the right solution, because it would not dismiss notifications on other connected devices like Android.
Current solution is to set a marker for the time when the code dismisses the existing notifications.

individual-it:
I've added more logic to remember the latest `updatedAt` field of all notifications associated with a WP, using that we don't have to dismiss all existing notifications and recreate them. This makes mobile clients only beep when something changed on the OP side

kiranparajuli589:
One case that I found:

### Steps to reproduce
1. Connect a normal user (not admin) from NC with the OP server.
2. Make notifications for that user from OP
3. Come to NC dashboard (login with the same normal user)
4. Open the NC notification dropdown and click the `(x)` icon.

### Expected
1. mark read should succeed in the NC
2. mark read should synchronize to OP

### Actual
1. mark read fails with `` in NC
    Request:
    ```json
    {
	    "DELETE": {
		    "scheme": "http",
		    "host": "localhost",
		    "filename": "/nextcloud/master/apps/integration_openproject/work-packages/4/notifications",
		    "remote": {
			    "Address": "127.0.0.1:80"
		    }
	    }
    }
    ```
    Response:
    ```json
    {
	    "message": "Logged in user must be an admin"
    }
    ```
3. OP detects no change

individual-it:
@kiranparajuli589 this is even a problem in the current master branch, fixed in #262 
Please test first there and after its merged I will rebase this PR

kiranparajuli589:
#262 merged.

individual-it:
@kiranparajuli589 I've rebased and pushed this branch

kiranparajuli589 (commented):
Success:
- **marking notification as read from the app** :
  - [x]  from the notifications panel of mobile: syncs with OP & NC web
  - [x] the notification center of the app itself: syncs with OP & NC web
- **there is existing unread notification, and a new is added for the same wp**
  - [x] new notification makes a beep on the phone...also syncs with NC web and OP web
- **there are existing unread notifications and cron job is rerun.**
  - [x] things are the way it was. 
One thing that was not working with me:
- **When I get notifications in NC and I mark them as read from the Dashboard widget...it gets' cleared from the OP but the NC notification are still there.**

individual-it:
@kiranparajuli589 I can mark notifications in the dashboard as read and they disappear, see attached video
can you please provide me steps to reproduce the issue

https://user-images.githubusercontent.com/2425577/200490929-eecbdae3-b742-433b-84b5-f9219ad91537.mp4

individual-it:
One problem that I've found is that when ALL notifications are marked as read the code does not execute, because I check if `count($notifications) > 0`. I will fix that

kiranparajuli589:
> @kiranparajuli589 I can mark notifications in the dashboard as read and they disappear, see attached video can you please provide me steps to reproduce the issue
I think I missed to run the `cron-job` after marking as read from the dashboard ;) everything works gr8. sorry for this inconvenience.

kiranparajuli589 (approved):

PR #257, opened 2022-10-18
[OP#44493] don't allow URLs that are redirected in the setup phase

individual-it:
In the setup phase URLs that are redirected should not be used. If an URL is redirected the admin should use the real URL

image

But if a sys-admin sets up a redirect after the configuration is done (e.g. for maintenance purposes) that redirect should be used and followed.

fixes https://community.openproject.org/work_packages/44493


kiranparajuli589 (approved):
lgtm :+1:

PR #259, opened 2022-10-26
 [OP#44482] pass in the OpenProject URL as state to the ProjectsTab

individual-it:
I don't see a reason why we should query the backend to find out the OpenProject URL, it does not change for different files or workpackage connections. It is much cheaper to pass it in by a state and using it.
Also it solves the problem with Firefox blocking the `window.open()` request as popup. Because the `window.open()` call now comes directly after the user interaction, Firefox does not block it.
For more details of how Firefox handles popups see: https://support.mozilla.org/en-US/kb/pop-blocker-settings-exceptions-troubleshooting?redirectslug=Pop-up+blocker&redirectlocale=en-US

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/44482

### DIFF
--- a/lib/Listener/LoadSidebarScript.php
+++ b/lib/Listener/LoadSidebarScript.php
@@ -103,5 +103,9 @@ class LoadSidebarScript implements IEventListener {
 		$this->initialStateService->provideInitialState(
 			'oauth-connection-error-message', $this->oauthConnectionErrorMessage
 		);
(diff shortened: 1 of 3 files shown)


kiranparajuli589 (approved):
lgtm :+1: checked with NC stable22-25 + latest master...works good everywhere.

PR #261, opened 2022-10-31
[OP#44727] invalidate oauth tokens given to OP

individual-it:
Currently when resetting the all integration settings in NC or changing the Nextcloud OAuth client, OpenProject can still use the acquired tokens, because deleting the oauth client does not invalidate existing tokens.

This is IMO a bug in Nextcloud and should be fixed there, see https://github.com/nextcloud/server/issues/35068 but before that is done we can use this workaround to invalidate all tokens associated with OpenProject in Nextcloud when deleting the oauth client.

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/44727

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -11,6 +11,7 @@
 
 namespace OCA\OpenProject\Controller;
 
+use OC\Authentication\Token\IProvider as IAuthTokenProvider;
 use OCP\IURLGenerator;
 use OCP\IConfig;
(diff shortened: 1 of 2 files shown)


kiranparajuli589 (commented):
lgtm :+1:

---

coverage % for unit tests is showing `0` https://github.com/nextcloud/integration_openproject/actions/runs/3360763036/jobs/5570349517

individual-it:
closing in favor of #273

PR #262, opened 2022-11-01
allow non admin user to mark notifications as read

individual-it:
without this change only admins are allowed to use the endpoint, this does not make sense and lead to https://github.com/nextcloud/integration_openproject/pull/256#issuecomment-1298094150


kiranparajuli589:
looks good. when `einstein` (non admin user) marks as read the notification from NC dashboard, the notification gets marked successfully in NC and synchronized in the OP. :+1:

kiranparajuli589:
CI should be green now :heavy_check_mark:

kiranparajuli589 (approved):

PR #270, opened 2022-11-09
[feature/44747] revoke client tokens received from openproject

kiranparajuli589:
Part of https://community.openproject.org/projects/nextcloud-integration/work_packages/44747

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -11,6 +11,7 @@
 
 namespace OCA\OpenProject\Controller;
 
+use GuzzleHttp\Exception\ConnectException;
 use OCA\OAuth2\Controller\SettingsController;
 use OCA\OAuth2\Exceptions\ClientNotFoundException;
 use OCP\IURLGenerator;
@@ -148,6 +149,7 @@ class ConfigController extends Controller {
 	 * @throws OpenprojectErrorException
 	 */
 	public function setAdminConfig(array $values): DataResponse {
+		$oPOAuthTokenRevokeStatus = "success";
 		$allowedKeys = [
 			'oauth_instance_url',
 			'client_id',
@@ -202,7 +204,7 @@ class ConfigController extends Controller {
 			($oldClientSecret && $oldClientId && $values['client_id'] === null && $values['client_secret'] === null)
 		) {
 			$this->userManager->callForAllUsers(function (IUser $user) use (
-				$oldOpenProjectOauthUrl, $oldClientId, $oldClientSecret
+				$oldOpenProjectOauthUrl, $oldClientId, $oldClientSecret, &$oPOAuthTokenRevokeStatus
 			) {
 				$userUID = $user->getUID();
 				$accessToken = $this->config->getUserValue($userUID, Application::APP_ID, 'token', '');
@@ -212,19 +214,35 @@ class ConfigController extends Controller {
 				// there may be cases where the software only have host url saved but not the client information
 				// in this case, the token is not revoked and just cleared if present in the database
 				if ($accessToken && $oldOpenProjectOauthUrl && $oldClientId && $oldClientSecret) {
-					$this->openprojectAPIService->revokeUserOAuthToken(
-						$oldOpenProjectOauthUrl,
-						$accessToken,
-						$oldClientId,
-						$oldClientSecret
-					);
+					try {
+						$this->openprojectAPIService->revokeUserOAuthToken(
+							$userUID,
+							$oldOpenProjectOauthUrl,
+							$accessToken,
+							$oldClientId,
+							$oldClientSecret
+						);
+					} catch (ConnectException $e) {
+						$oPOAuthTokenRevokeStatus = "connection_error";
+						$this->logger->error(
+							'Error: ' . $e->getMessage(),
+							['app' => Application::APP_ID]
+						);
+					} catch (OpenprojectErrorException $e) {
+						$oPOAuthTokenRevokeStatus = "other_error";
+						$this->logger->error(
+							'Error: ' . $e->getMessage(),
+							['app' => Application::APP_ID]
+						);
+					}
 				}
 				$this->clearUserInfo($userUID);
 			});
 		}
 
 		return new DataResponse([
-			"status" => OpenProjectAPIService::isAdminConfigOk($this->config)
+			"status" => OpenProjectAPIService::isAdminConfigOk($this->config),
+			"oPOAuthTokenRevokeStatus" => $oPOAuthTokenRevokeStatus
 		]);
 	}
 
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -27,6 +27,7 @@ use OCP\AppFramework\Controller;
 use OCA\OpenProject\Service\OauthService;
 use OCA\OpenProject\Service\OpenProjectAPIService;
 use OCA\OpenProject\AppInfo\Application;
+use OCA\OpenProject\Exception\OpenprojectErrorException;
 use Psr\Log\LoggerInterface;
 
 class ConfigController extends Controller {
@@ -144,6 +145,7 @@ class ConfigController extends Controller {
 	 * @param array<string, string|null> $values
 	 *
 	 * @return DataResponse
+	 * @throws OpenprojectErrorException
 	 */
 	public function setAdminConfig(array $values): DataResponse {
 		$allowedKeys = [
@@ -151,7 +153,8 @@ class ConfigController extends Controller {
 			'client_id',
 			'client_secret',
 			'default_enable_navigation',
-			'default_enable_unified_search'
+			'default_enable_unified_search',
+			'default_enable_notifications'
 		];
 
 		// if values contains a key that is not in the allowedKeys array,
@@ -191,10 +194,32 @@ class ConfigController extends Controller {
 				);
 			}
 		}
-		if ((key_exists('client_id', $values) && $values['client_id'] !== $oldClientId) ||
-			(key_exists('client_secret', $values) && $values['client_secret'] !== $oldClientSecret)) {
-			$this->userManager->callForAllUsers(function (IUser $user) {
-				$this->clearUserInfo($user->getUID());
+		if (
+			// when the OP client information has changed
+			((key_exists('client_id', $values) && $values['client_id'] !== $oldClientId) ||
+			(key_exists('client_secret', $values) && $values['client_secret'] !== $oldClientSecret)) ||
+			// when the OP client information is for reset
+			($oldClientSecret && $oldClientId && $values['client_id'] === null && $values['client_secret'] === null)
+		) {
+			$this->userManager->callForAllUsers(function (IUser $user) use (
+				$oldOpenProjectOauthUrl, $oldClientId, $oldClientSecret
+			) {
+				$userUID = $user->getUID();
+				$accessToken = $this->config->getUserValue($userUID, Application::APP_ID, 'token', '');
+
+				// revoke is possible only when the user has the access token stored in the database
+				// plus, for a successful revoke, the old OP client information is also needed
+				// there may be cases where the software only have host url saved but not the client information
+				// in this case, the token is not revoked and just cleared if present in the database
+				if ($accessToken && $oldOpenProjectOauthUrl && $oldClientId && $oldClientSecret) {
+					$this->openprojectAPIService->revokeUserOAuthToken(
+						$oldOpenProjectOauthUrl,
+						$accessToken,
+						$oldClientId,
+						$oldClientSecret
+					);
+				}
+				$this->clearUserInfo($userUID);
 			});
 		}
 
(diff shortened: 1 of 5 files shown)


lib/Controller/ConfigController.php
```diff
@@ -364,4 +379,33 @@ public function checkConfig(): DataResponse {
...
+	 */
+	private function revokeUserToken(
+		string $opUrl,
+		string $userId,
+		string $opOAuthClientID,
+		string $opOAuthClientSecret
+	): void {
+		$accessToken = $this->config->getUserValue($userId, Application::APP_ID, 'token', '');
+
+		if ($opUrl && $opOAuthClientID && $opOAuthClientSecret && $accessToken) {
```
individual-it, line 402: what is that expression is not true, then the code will silently do nothing. I think its better to throw an exception

lib/Service/OpenProjectAPIService.php
```diff
@@ -879,4 +879,39 @@ public static function sanitizeUrl(
...
+	 *
+	 * @param string $openProjectUrl the url of the openproject instance
+	 * @param string $accessToken the refresh token to be revoked
+	 * @param string $clientId the client id of the OAuth app
+	 * @param string $clientSecret the client secret of the OAuth app
+	 *
+	 * @return void
+	 * @throws OpenprojectErrorException
+	 */
+	public function revokeUserOAuthToken(
```
individual-it, line 894: why not integrate that function directly in `revokeUserToken`?
I think the most other functions in this file are providing some REST API, but this one not

lib/Controller/ConfigController.php
```diff
@@ -184,9 +186,22 @@ public function setAdminConfig(array $values): DataResponse {
...
 		}
-		if ((key_exists('client_id', $values) && $values['client_id'] !== $oldClientId) ||
-			(key_exists('client_secret', $values) && $values['client_secret'] !== $oldClientSecret)) {
-			$this->userManager->callForAllUsers(function (IUser $user) {
+		if (
+			// when the OP client information has changed
+			((key_exists('client_id', $values) && $values['client_id'] !== $oldClientId) ||
+			(key_exists('client_secret', $values) && $values['client_secret'] !== $oldClientSecret)) ||
+			// when the OP client information is for reset
+			($oldClientSecret && $oldClientId && $values['client_id'] === null && $values['client_secret'] === null)
```
individual-it, line 194: is that a bug in the current code? On reset we don't clean the user info?

individual-it (changes requested):

lib/Service/OpenProjectAPIService.php
kiranparajuli589: This func. sends a `POST` request for the revoke action to the OP instance. So I had put it in the `OpenProjectAPIService` section. We do NC things in the `ConfigController`.  Should it be moved? I can also call _revoke_ directly from the `setAdminConfig` fn.

lib/Controller/ConfigController.php
kiranparajuli589: maybe yes :)

lib/Controller/ConfigController.php
kiranparajuli589: there can be case where user might not have the `accessToken` set. I've modified the code. Please review.

lib/Service/OpenProjectAPIService.php
individual-it: I think I would put it into `revokeUserToken`

lib/Service/OpenProjectAPIService.php
```diff
@@ -879,4 +879,39 @@ public static function sanitizeUrl(
...
+			]
+		];
+		$response = $this->client->post(
+			$openProjectUrl . '/oauth/revoke' . '?token=' . $accessToken,
+			$options,
+		);
+		$body = $response->getBody();
+		$respCode = $response->getStatusCode();
+		if ($respCode !== 200) {
+			throw new OpenprojectErrorException($body);
```
individual-it, line 914:
```suggestion
			throw new OpenprojectErrorException('could not revoke token in OpenProject. response body: "' .$body . '"');
```
what will happen in the UI in that case? Will the user see some technical error?
Maybe better to have some nice error message to the user and a detailed log entry

individual-it (changes requested):

SwikritiT:
@kiranparajuli589 Seems like the CI error is because of the passing variable by reference here https://github.com/nextcloud/integration_openproject/pull/270/commits/da3cc2af9fac34c318fbe67580990870139b8197 
If I pass them as value CI is green https://github.com/nextcloud/integration_openproject/actions/runs/3628009178, but the feature won't work as intended?

lib/Service/OpenProjectAPIService.php
```diff
@@ -725,4 +725,39 @@ public static function sanitizeUrl(
...
+		];
+		$response = $this->client->post(
+			$openProjectUrl . '/oauth/revoke' . '?token=' . $accessToken,
+			$options,
+		);
+		$body = $response->getBody();
+		$respCode = $response->getStatusCode();
+		if ($respCode !== 200) {
+			throw new OpenprojectErrorException('Could not revoke token in OpenProject. Response body: "' .$body . '"');
+		}
```
individual-it, line 761: if the OP server cannot be reached the user does not see any error message and the log-file only shows an exception like 
```
  "reqId": "w1EeF0qP7dEfEnJSosrG",
  "level": 3,
  "time": "2022-12-07T05:22:14+00:00",
  "remoteAddr": "127.0.0.1",
  "user": "admin",
  "app": "index",
  "method": "PUT",
  "url": "/nextcloud-server/index.php/apps/integration_openproject/admin-config",
  "message": "cURL error 7: Failed to connect to host.docker.internal port 3000 after 0 ms: Connection refused (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://host.docker.internal:3000/oauth/revoke?token=sLlQ4LHmS_jC1IW8SS8M-1edpOB72m0-AvtPb9sPDJE",
  "userAgent": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:107.0) Gecko/20100101 Firefox/107.0",
  "version": "24.0.7.1",
  "exception": {
    "Exception": "GuzzleHttp\\Exception\\ConnectException",
    "Message": "cURL error 7: Failed to connect to host.docker.internal port 3000 after 0 ms: Connection refused (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://host.docker.internal:3000/oauth/revoke?token=sLlQ4LHmS_jC1IW8SS8M-1edpOB72m0-AvtPb9sPDJE",
    "Code": 0,
    "Trace": [
      {
        "file": "/home/artur/www/nextcloud-server/3rdparty/guzzlehttp/guzzle/src/Handler/CurlFactory.php",
        "line": 158,
        "function": "createRejection",
        "class": "GuzzleHttp\\Handler\\CurlFactory",
        "type": "::"
      },
      {
        "file": "/home/artur/www/nextcloud-server/3rdparty/guzzlehttp/guzzle/src/Handler/CurlFactory.php",
```

I think it would be better to show a general message to the user like `OAuth Tokens in OpenProject could not be reset` and a good log message with more details like response code and body

lib/Controller/ConfigController.php
```diff
@@ -191,10 +193,33 @@ public function setAdminConfig(array $values): DataResponse {
...
+		) {
+			$this->userManager->callForAllUsers(function (IUser $user) use (
+				$oldOpenProjectOauthUrl, $oldClientId, $oldClientSecret
+			) {
+				$userUID = $user->getUID();
+				if (!$oldOpenProjectOauthUrl || !$oldClientId || !$oldClientSecret) {
+					throw new Exception('Missing required parameters.'
```
individual-it, line 208: do we need to throw an exception here?
maybe better to show a message to the user and have a log entry

e.g. when the admin enters some wrong OP client id and secret at the first time setting up the app and then clicks on "Replace OpenProject OAuth values" only the message `Missing  required parameters.` is shown and the admin cannot replace the values

individual-it (changes requested):

lib/Controller/ConfigController.php
```diff
@@ -191,10 +193,33 @@ public function setAdminConfig(array $values): DataResponse {
...
-		if ((key_exists('client_id', $values) && $values['client_id'] !== $oldClientId) ||
-			(key_exists('client_secret', $values) && $values['client_secret'] !== $oldClientSecret)) {
-			$this->userManager->callForAllUsers(function (IUser $user) {
-				$this->clearUserInfo($user->getUID());
+		if (
+			// when the OP client information has changed
+			((key_exists('client_id', $values) && $values['client_id'] !== $oldClientId) ||
+			(key_exists('client_secret', $values) && $values['client_secret'] !== $oldClientSecret)) ||
+			// when the OP client information is for reset
+			($oldClientSecret && $oldClientId && $values['client_id'] === null && $values['client_secret'] === null)
```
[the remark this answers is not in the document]
kiranparajuli589: for that case, revoke would not be possible. if we have the token present for the user, it will be cleared in the loop below. any more thing we should do here @individual-it ?

lib/Service/OpenProjectAPIService.php
kiranparajuli589: should we clear the user info & admin config data, if the revoke fails?

individual-it: I think yes

lib/Controller/ConfigController.php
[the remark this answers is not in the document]
individual-it: lets tell the user that revoking was not possible, but still reset

lib/Service/OpenProjectAPIService.php
kiranparajuli589: upto now...the admin data is cleared at first, then in user for loop: revoke is done and then user data is cleaned up

kiranparajuli589:
unit tests for UI and Pact tests for service API is yet to be added.

individual-it (changes requested):

lib/Controller/ConfigController.php
```diff
@@ -191,15 +195,53 @@ public function setAdminConfig(array $values): DataResponse {
...
+		) {
+			$this->userManager->callForAllUsers(function (IUser $user) use (
+				$oldOpenProjectOauthUrl, $oldClientId, $oldClientSecret, &$oPOAuthTokenRevokeStatus
```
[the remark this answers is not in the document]
kiranparajuli589: the value of `$oPOAuthTokenRevokeStatus` has to be changed inside the callback fn. which is only possible by using the reference. i am trying to see why this coverage is failing. suggestions?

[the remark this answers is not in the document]
kiranparajuli589: done using the system variables

lib/Controller/ConfigController.php
```diff
@@ -144,14 +145,17 @@ public function setConfig(array $values): DataResponse {
...
 	public function setAdminConfig(array $values): DataResponse {
+		$oPOAuthTokenRevokeStatus = "success";
 		$allowedKeys = [
 			'oauth_instance_url',
 			'client_id',
 			'client_secret',
 			'default_enable_navigation',
-			'default_enable_unified_search'
+			'default_enable_unified_search',
+			'default_enable_notifications'
```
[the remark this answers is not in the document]
kiranparajuli589: fixed

individual-it (changes requested):
there are still problems:
1. with a full setup
2. try to change the OP client id & secret
=> the UI reloads and leaves the id & secrets empty :-1: 

1. reset everything
4. enter valid openproject URL
 => UI says the tokens have been revoked :-1: because there haven't been any tokens to revoke
 => the UI does not go to the next part :-1:

SwikritiT:
> there are still problems:
> 
> 1. with a full setup
> 2. try to change the OP client id & secret
>    => the UI reloads and leaves the id & secrets empty -1
> 3. reset everything
> 4. enter valid openproject URL
>    => UI says the tokens have been revoked -1 because there haven't been any tokens to revoke
>    => the UI does not go to the next part -1

I cannot reproduce this now, @kiranparajuli589 but some of the php tests are still failing

kiranparajuli589:
@SwikritiT I've fixed the unit tests too. please give it a look again.

SwikritiT:
I'm having this problem now
I cannot connect through the dashboard of the normal user or rather I can connect but it still shows the connect button
Screenshot from 2022-12-09 17-15-42


Screenshot from 2022-12-09 17-15-54

Not reproducible for admin user though

lib/Controller/ConfigController.php
```diff
@@ -191,15 +197,59 @@ public function setAdminConfig(array $values): DataResponse {
 				);
 			}
 		}
-		if ((key_exists('client_id', $values) && $values['client_id'] !== $oldClientId) ||
-			(key_exists('client_secret', $values) && $values['client_secret'] !== $oldClientSecret)) {
-			$this->userManager->callForAllUsers(function (IUser $user) {
-				$this->clearUserInfo($user->getUID());
+		if (
```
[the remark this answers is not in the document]
kiranparajuli589: done

kiranparajuli589:
@SwikritiT I have tried and it works fine with stable 24 and 25. Can u please check in with NC master branch.

SwikritiT:
looks good

individual-it (approved):

PR #272, opened 2022-11-14
[OP#44485]Enable Reset Button only when any integration setting is set by admin

SagarGi:
### Description
During `Oauth` configuration between 2 apps, even when admin has not configured any setting the `reset` button is enabled. This PR disable it until any of the configuration is set.

### Related issue and fixes:
https://community.openproject.org/projects/nextcloud-integration/work_packages/44485/activity?query_id=3513

UI Before changes: When no setting is set by admin
Screenshot from 2022-11-14 16-30-36

UI After changes: When a setting is set by admin

Screenshot from 2022-11-14 16-30-48

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -163,7 +163,7 @@
 		</div>
 		<Button id="reset-all-app-settings-btn"
 			type="error"
-			:disabled="!isIntegrationComplete"
+			:disabled="isResetButtonDisabled"
 			@click="resetAllAppValuesConfirmation">
 			<template #icon>
 				<RestoreIcon :size="20" />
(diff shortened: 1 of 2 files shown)


SwikritiT:
The changes looks good, maybe we can add some unit tests as well?

src/components/AdminSettings.vue
```diff
@@ -163,6 +163,7 @@
 		</div>
 		<Button id="reset-all-app-settings-btn"
 			type="error"
+			:disabled="!isServerHostFormComplete"
```
kiranparajuli589, line 166: what if we clear the OP server host form but we still have OP Client Information? will such case exists, I'm not sure :) the the reset button would be disabled on that case with this but we still have something to clear.

individual-it:
@SagarGi please sign your commits e.g. with `git commit -m"...." -s`

kiranparajuli589 (approved):

individual-it (approved):

PR #273, opened 2022-11-15
[OP#44727] let NC delete the OAuth client

individual-it:
The current code does the oauth client deletion the same way as it is done in the NC code
compare https://github.com/nextcloud/integration_openproject/blob/master/lib/Service/OauthService.php#L113-L119 with https://github.com/nextcloud/server/blob/master/apps/oauth2/lib/Controller/SettingsController.php#L93-L96

This PR lets the NC code handle the deleting and together with https://github.com/nextcloud/server/pull/35094 it is an alternative to #261 and will fix https://community.openproject.org/projects/nextcloud-integration/work_packages/44727

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -11,6 +11,8 @@
 
 namespace OCA\OpenProject\Controller;
 
+use OCA\OAuth2\Controller\SettingsController;
+use OCA\OAuth2\Exceptions\ClientNotFoundException;
 use OCP\IURLGenerator;
 use OCP\IConfig;
 use OCP\IL10N;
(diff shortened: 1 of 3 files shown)


kiranparajuli589 (approved):
- [x] NC OAuth is successfully removed while restting the admin configurations
- [x] NC OAuth is successfully updated after updating the `oauth_instance_url` in the Server Host Form

LGTM :+1:

individual-it:
I believe we can merge this. When https://github.com/nextcloud/server/pull/35094 gets into the NC server core it will change details what happens on client deletion, but that should not affect us here except of solving the original issue

PR #279, opened 2022-11-24
[OP#42323] support dateAlert notifications from OP

individual-it:
from the version 12.4 OpenProject sends out notifications based on the dates set in the WP. This PR supports those notifications now. Before date notifications broke the dashboard and would display wrongly in the notifications app, because they don't have a actor.

For now the dashboard shows a `?` if there is no actor, after #254 we can use a calendar icon for that

To test date based notifications this command can be used to generate them :
`rails runner 'Notification.create(recipient: User.find_by(lastname: "[LASTNAME OF USER TO CREATE NOTIFICATION FOR]"), reason: :[REASON], resource: WorkPackage.find([SOME ID]), project: WorkPackage.find([SOME SAME ID]).project)'`
valid values for `reason` are: `date_alert_start_date`, `date_alert_due_date` and `date_alert_date`

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -183,16 +183,19 @@ export default {
 						const userId = n._links?.actor?.href
 							? n._links.actor.href.replace(/.*\//, '')
 							: null
-						if (notifications[wpId].mostRecentActor === undefined) {
+						const title = n._links?.actor?.title
+							? n._links.actor.title
+							: null
+						if (notifications[wpId].mostRecentActor === undefined && userId !== null) {
 							notifications[wpId].mostRecentActor = {
-								title: n._links.actor.title,
+								title,
 								id: userId,
 								createdAt: n.createdAt,
 							}
-						} else if (userId !== notifications[wpId].mostRecentActor.id) {
+						} else if (userId !== null && userId !== notifications[wpId].mostRecentActor.id) {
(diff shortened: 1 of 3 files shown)


individual-it:
converted back to draft as it needs to be rebased on #282, some code will disappear from here

individual-it:
ready for review again

src/views/Dashboard.vue
```diff
@@ -183,16 +183,19 @@ export default {
 						const userId = n._links?.actor?.href
 							? n._links.actor.href.replace(/.*\//, '')
 							: null
-						if (notifications[wpId].mostRecentActor === undefined) {
+						const title = n._links?.actor?.title
+							? n._links.actor.title
+							: null
```
kiranparajuli589, line 188:
```suggestion
						const title = n._links?.actor?.title || null
```

kiranparajuli589 (commented):
code lgtm :+1:

src/views/Dashboard.vue
individual-it: I've been just following the format of the lines above

SwikritiT:
Looks good 

Screenshot from 2022-12-07 17-39-51

Screenshot from 2022-12-07 17-40-06

SwikritiT (approved):

PR #280, opened 2022-11-28
[OP#44964]Single API endpoint for setting whole integration

SagarGi:
### Description
This PR adds an endpoint for setting of whole integration configuration.

### Related Issue
https://community.openproject.org/projects/nextcloud-integration/work_packages/44964/activity?query_id=3513

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,66 +22,71 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
 - For documentation on how to set up the integration as an administrator, refer to [Nextcloud integration setup](https://openproject.org/docs/system-admin-guide/integrations/nextcloud/).
 
 ## Setting up the integration as an administrator with API
-We have a single API endpoint (/setup) available to set , update and delete the integration.
-1. Set up the whole integration with [POST] request
-We need to provide following open project key information in data
-	- openproject_instance_url
-    - openproject_client_id
-    - openproject_client_id
+We have a single API endpoint (/setup) available to set up, update and reset the integration.
 
-Example curl request to set up whole integration
-```bash
-curl -XPOST -uadmin:admin  http://<nextcloud_host>/index.php/apps/integration_openproject/setup \
--d '{"values":{"openproject_instance_url":<openproject_instance_url>,"openproject_client_id":<openproject_client_id>,"openproject_client_secret":<openproject_client_secret>,"default_enable_navigation":null,"default_enable_unified_search":null}}' \
--H 'Content-Type: application/json' -v
-```
+To set up or update the integration following data needs to be provided:
+- openproject_instance_url
+- openproject_client_id
+- openproject_client_id
+- default_enable_navigation
+- default_enable_unified_search
 
-The response from the above curl request
-```json
-{
-    "nextcloud_oauth_client_name": <openproject-client>,
-    "openproject_redirect_uri": "http://<openproject_instance_url>/oauth_clients/<nextcloud_client_id>/callback",
-    "nextcloud_client_id": <nextcloud_client_id>,
-    "nextcloud_client_secret": <nextcloud_client_secret>
-}
-```
 
-2. Update the integration with [PUT] request
-   We can provide any of following open project key information in data to update
-	- openproject_instance_url
-	- openproject_client_id
-	- openproject_client_id
+1. **Set up the whole integration with a [POST] request**
 
-Example curl request to update only `openproject_client_id` and `openproject_client_secret`
-```bash
-curl -XPUT -uadmin:admin  http://<nextcloud_host>/index.php/apps/integration_openproject/setup \
--d '{"values":{"openproject_client_id":<openproject_client_id>,"openproject_client_secret":<openproject_client_secret>}}' \
--H 'Content-Type: application/json' -v
-```
+	We must provide all of the above data for this request.
 
-The response from the above curl request
-```json
-{
-    "nextcloud_oauth_client_name": <openproject-client>,
-    "openproject_redirect_uri": "http://<openproject_instance_url>/oauth_clients/<nextcloud_client_id>/callback",
-    "nextcloud_client_id": <nextcloud_client_id>,
-    "nextcloud_client_secret": <nextcloud_client_secret>
-}
-```
+	Example curl request to set up whole integration
+	```bash
+	curl -XPOST -uadmin:admin  http://<nextcloud_host>/index.php/apps/integration_openproject/setup \
+	-d '{"values":{"openproject_instance_url":<openproject_instance_url>,"openproject_client_id":<openproject_client_id>,"openproject_client_secret":<openproject_client_secret>,"default_enable_navigation":false,"default_enable_unified_search":false}}' \
+	-H 'Content-Type: application/json' -v
+	```
 
-2. Resetting the whole integration with [DELETE] request
+	The response from the above curl request
+	```json
+	{
+		"nextcloud_oauth_client_name": <openproject-client>,
+		"openproject_redirect_uri": "http://<openproject_instance_url>/oauth_clients/<nextcloud_client_id>/callback",
+		"nextcloud_client_id": <nextcloud_client_id>,
+		"nextcloud_client_secret": <nextcloud_client_secret>
+	}
+	```
 
-Example curl request to reset whole integration
-```bash
-curl -XDELETE -uadmin:admin http://<nextcloud_host>/index.php/apps/integration_openproject/setup
-```
+2. **Update the integration with a [PUT] request**
 
-The response from the above curl request
-```json
-{
-	"status": true
-}
-```
+    One or more of the above data needs to be sent to this endpoint
+
+	Example curl request to update only `openproject_client_id` and `openproject_client_secret`
+	```bash
+	curl -XPUT -uadmin:admin  http://<nextcloud_host>/index.php/apps/integration_openproject/setup \
+	-d '{"values":{"openproject_client_id":<openproject_client_id>,"openproject_client_secret":<openproject_client_secret>}}' \
+	-H 'Content-Type: application/json' -v
+	```
+
+	The response from the above curl request
+	```json
+	{
+		"nextcloud_oauth_client_name": <openproject-client>,
+		"openproject_redirect_uri": "http://<openproject_instance_url>/oauth_clients/<nextcloud_client_id>/callback",
+		"nextcloud_client_id": <nextcloud_client_id>,
+		"nextcloud_client_secret": <nextcloud_client_secret>
+	}
+	```
+
+2. **Resetting the whole integration with a [DELETE] request**
+
+	Example curl request to reset whole integration
+	```bash
+	curl -XDELETE -uadmin:admin http://<nextcloud_host>/index.php/apps/integration_openproject/setup
+	```
+
+	The response from the above curl request
+	```json
+	{
+		"status": true
+	}
+	```
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -21,6 +21,68 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
 - For documentation on how to use the integration once it is set up, refer to [Using the Nextcloud integration](https://openproject.org/docs/user-guide/nextcloud-integration/).
 - For documentation on how to set up the integration as an administrator, refer to [Nextcloud integration setup](https://openproject.org/docs/system-admin-guide/integrations/nextcloud/).
 
+## Setting up the integration as an administrator with API
+We have a single API endpoint (/setup) available to set , update and delete the integration.
+1. Set up the whole integration with [POST] request
+We need to provide following open project key information in data
+	- openproject_instance_url
+    - openproject_client_id
+    - openproject_client_id
+
+Example curl request to set up whole integration
+```bash
+curl -XPOST -uadmin:admin  http://<nextcloud_host>/index.php/apps/integration_openproject/setup \
+-d '{"values":{"openproject_instance_url":<openproject_instance_url>,"openproject_client_id":<openproject_client_id>,"openproject_client_secret":<openproject_client_secret>,"default_enable_navigation":null,"default_enable_unified_search":null}}' \
+-H 'Content-Type: application/json' -v
+```
+
+The response from the above curl request
+```json
+{
+    "nextcloud_oauth_client_name": <openproject-client>,
+    "openproject_redirect_uri": "http://<openproject_instance_url>/oauth_clients/<nextcloud_client_id>/callback",
+    "nextcloud_client_id": <nextcloud_client_id>,
+    "nextcloud_client_secret": <nextcloud_client_secret>
+}
+```
+
+2. Update the integration with [PUT] request
+   We can provide any of following open project key information in data to update
+	- openproject_instance_url
+	- openproject_client_id
+	- openproject_client_id
+
+Example curl request to update only `openproject_client_id` and `openproject_client_secret`
+```bash
+curl -XPUT -uadmin:admin  http://<nextcloud_host>/index.php/apps/integration_openproject/setup \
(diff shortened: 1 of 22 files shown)


lib/Controller/ConfigController.php
```diff
@@ -196,6 +196,26 @@ public function setAdminConfig(array $values): DataResponse {
...
+	 */
+	public function setUpWholeIntegration(array $values): DataResponse {
+		$responseAdminConfigSet = $this->setAdminConfig($values);
+		if ($responseAdminConfigSet->getData()['status']) {
+			$responseFromOAuthCreation = $this->autoOauthCreation();
+			return new DataResponse([
+				"nextcloud_admin_config_status" => $responseAdminConfigSet->getData()['status'],
+				"openproject_client_info" => $responseFromOAuthCreation->getData()
+			]);
+		}
```
kiranparajuli589, line 215: else some error need to be thrown? user also should be shown some message about the failure.

kiranparajuli589: on the second thought,,,,server log may be useful but this is not yet used by the UI. ur call.

SwikritiT (commented):
maybe we can add some PHP unit tests as well?

SwikritiT:
sending the request with following data 
```json
{
		"values":
			{
				"client_id":null,
				"client_secret":null,
				"default_enable_navigation":false,
				"default_enable_unified_search":false,
				"oauth_instance_url":null
			}
}
```
GIves me `500` maybe it should support `null` as well, while resetting we sent null value

SwikritiT:
In another case while sending `empty`, `client-id`, and `client_secret` it sends me `200`  and the UI looks like this
Screenshot from 2022-12-07 15-22-55


Maybe a check for empty string?

lib/Controller/ConfigController.php
```diff
@@ -109,58 +109,28 @@ public function clearUserInfo(string $userId = null) {
...
 		// return a response with status code 400 and an error message
 		foreach ($values as $key => $value) {
 			if (!in_array($key, $allowedKeys)) {
-				return new DataResponse([
-					'error' => $this->l->t('Invalid key')
-				], Http::STATUS_BAD_REQUEST);
+				return [
+					'response_type' => 'error',
+					'response_result' => $this->l->t('Invalid key'),
+					'response_status_code' => Http::STATUS_BAD_REQUEST
```
kiranparajuli589, line 132:
```suggestion
					'type' => 'error',
					'result' => $this->l->t('Invalid key'),
					'status_code' => Http::STATUS_BAD_REQUEST
```

lib/Controller/ConfigController.php
```diff
@@ -109,58 +109,28 @@ public function clearUserInfo(string $userId = null) {
...
-	 * set admin config values
+	 * set up open-project admin setup
 	 *
 	 * @param array<string, string|null> $values
-	 *
-	 * @return DataResponse
+	 * @return array<mixed>
 	 */
-	public function setAdminConfig(array $values): DataResponse {
+	private function setUpOpenProjectAdminSetUp(array $values): array {
```
kiranparajuli589, line 117:
```suggestion
	private function setUpOpenProjectAdminConfig(array $values): array {
```

lib/Controller/ConfigController.php
```diff
@@ -197,10 +164,128 @@ public function setAdminConfig(array $values): DataResponse {
...
+		$mustHaveClientKey = [
+			'oauth_instance_url',
+			'client_id',
+			'client_secret',
+		];
+
+		foreach ($mustHaveClientKey as $keyName) {
+			if (!array_key_exists($keyName, $values)) {
+				$responseToSend['op_setup_info'] = [
+					"status" => "Missing Open Project client info or Invalid Key"
```
individual-it, line 252:
```suggestion
					"status" => "Missing OpenProject client information or invalid key"
```

lib/Controller/ConfigController.php
```diff
@@ -197,10 +164,128 @@ public function setAdminConfig(array $values): DataResponse {
...
+				];
+				return new DataResponse($responseToSend, HTTP::STATUS_BAD_REQUEST);
+			}
+			if ($keyName == "default_enable_navigation" || $keyName == "default_enable_unified_search") {
+				$values[$keyName] = (bool)$values[$keyName];
+			}
+			$values[$keyName] = (string)$values[$keyName];
```
individual-it, line 259:
```suggestion
			if ($keyName == "default_enable_navigation" || $keyName == "default_enable_unified_search") {
				$values[$keyName] = (bool)$values[$keyName];
			} else {
			        $values[$keyName] = (string)$values[$keyName];
		        }
```
otherwise it will be cast to bool and then to string

lib/Controller/ConfigController.php
```diff
@@ -197,10 +164,128 @@ public function setAdminConfig(array $values): DataResponse {
...
 			});
 		}
+	}
 
+	/**
+	 * returns auto-generated OAuth client information for nextcloud
+	 *
+	 * @return array<mixed>
+	 */
+	private function createAndGetClientInformationForNextCloud(): array {
```
individual-it, line 174:
```suggestion
	private function recreateOAuthClient(): array {
```

lib/Controller/ConfigController.php
```diff
@@ -197,10 +164,128 @@ public function setAdminConfig(array $values): DataResponse {
 				$this->clearUserInfo($user->getUID());
 			});
 		}
+	}
 
+	/**
+	 * returns auto-generated OAuth client information for nextcloud
```
individual-it, line 170:
```suggestion
	 * recreates the OAuth client and returns its details
```

appinfo/routes.php
```diff
@@ -14,6 +14,7 @@
 		['name' => 'config#oauthRedirect', 'url' => '/oauth-redirect', 'verb' => 'GET'],
 		['name' => 'config#setConfig', 'url' => '/config', 'verb' => 'PUT'],
 		['name' => 'config#setAdminConfig', 'url' => '/admin-config', 'verb' => 'PUT'],
+		['name' => 'config#setUpWholeIntegration', 'url' => '/setup-whole-integration', 'verb' => 'PUT'],
```
individual-it, line 17:
```suggestion
		['name' => 'config#setup', 'url' => '/setup', 'verb' => 'PUT'],
```

individual-it (changes requested):

lib/Controller/ConfigController.php
```diff
@@ -374,4 +370,123 @@ public function checkConfig(): DataResponse {
...
+			'default_enable_unified_search'
+		];
+		// check for allowed keys only
+		foreach ($values as $key => $value) {
+			if (!in_array($key, $allowedKeys)) {
+				return new DataResponse([
+					'error' => $this->l->t('invalid key')
+				], Http::STATUS_BAD_REQUEST);
+			}
+		}
```
kiranparajuli589, line 439: this check can be reused with some separate fn.

lib/Controller/ConfigController.php
```diff
@@ -374,4 +370,123 @@ public function checkConfig(): DataResponse {
...
+			"default_enable_navigation",
+			"default_enable_unified_search",
+		];
+		foreach ($mustHaveKey as $key) {
+			$this->config->setAppValue(Application::APP_ID, $key, '');
+		}
+		// also delete oAuthClient
+		$this->deleteOauthClient();
+		return new DataResponse([
+			"message" => "Delete Successful"
```
kiranparajuli589, line 476:
```suggestion
			"message" => "Reset Successful"
```

lib/Service/OpenProjectAPIService.php
```diff
@@ -453,6 +453,34 @@ public static function validateURL(string $url): bool {
...
+	 * validates the provided data for integration setup
+	 *
+	 * @param array<mixed> $values
+	 * @return bool
+	 * @throws \InvalidArgumentException
+	 *
+	 */
+	public static function validateIntegrationSetupInformation(array $values): bool {
+		foreach ($values as $key => $value) {
+			if ($key == 'openproject_instance_url' && !OpenProjectAPIService::validateURL((string)$value)) {
```
individual-it, line 466: always use `===` to check also the type
```suggestion
			if ($key === 'openproject_instance_url' && !OpenProjectAPIService::validateURL((string)$value)) {
```
why casting `$value` to string?
shouldn't `validateURL` return false if the parameter is not string?

lib/Service/OpenProjectAPIService.php
```diff
@@ -453,6 +453,34 @@ public static function validateURL(string $url): bool {
...
+				return false;
+			}
+			// validating specific two key
+			if ($key == 'default_enable_navigation' || $key == 'default_enable_unified_search') {
+				if (!is_bool($value)) {
```
individual-it, line 471:
```suggestion
			if (($key === 'default_enable_navigation' || $key === 'default_enable_unified_search') && !is_bool($value)) {
```

individual-it (changes requested):

individual-it (changes requested):
sending invalid JSON results in error 500, see tests added in https://github.com/nextcloud/integration_openproject/pull/280/commits/428e05391fdd89cea91a266ce4ba86f2d2e49d0e

lib/Service/OpenProjectAPIService.php
SagarGi: i do not want to run the code below after this condition so i have checked it with extra if since i am using `continue`.

README.md
```diff
@@ -21,6 +21,68 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
 - For documentation on how to use the integration once it is set up, refer to [Using the Nextcloud integration](https://openproject.org/docs/user-guide/nextcloud-integration/).
 - For documentation on how to set up the integration as an administrator, refer to [Nextcloud integration setup](https://openproject.org/docs/system-admin-guide/integrations/nextcloud/).
 
+## Setting up the integration as an administrator with API
+We have a single API endpoint (/setup) available to set , update and delete the integration.
```
individual-it, line 25:
```suggestion
We have a single API endpoint (/setup) available to set up, update and delete the integration.
```

README.md
```diff
@@ -21,6 +21,68 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
...
+1. Set up the whole integration with [POST] request
+We need to provide following open project key information in data
+	- openproject_instance_url
+    - openproject_client_id
```
individual-it, line 29:
```suggestion
These data need to be provided:
    - openproject_instance_url
    - openproject_client_id
```

README.md
```diff
@@ -21,6 +21,68 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
...
+    - openproject_client_id
+
+Example curl request to set up whole integration
+```bash
+curl -XPOST -uadmin:admin  http://<nextcloud_host>/index.php/apps/integration_openproject/setup \
+-d '{"values":{"openproject_instance_url":<openproject_instance_url>,"openproject_client_id":<openproject_client_id>,"openproject_client_secret":<openproject_client_secret>,"default_enable_navigation":null,"default_enable_unified_search":null}}' \
```
individual-it, line 35: I think default_enable_navigation and default_enable_unified_search need to be "true" or "false"

README.md
```diff
@@ -21,6 +21,68 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
...
+    "nextcloud_client_id": <nextcloud_client_id>,
+    "nextcloud_client_secret": <nextcloud_client_secret>
+}
+```
+
+2. Update the integration with [PUT] request
+   We can provide any of following open project key information in data to update
+	- openproject_instance_url
+	- openproject_client_id
+	- openproject_client_id
```
individual-it, line 53:
```suggestion
One or more of this items need to be send to the endpoint
    - openproject_instance_url
    - openproject_client_id
    - openproject_client_id
```

README.md
```diff
@@ -21,6 +21,68 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
 - For documentation on how to use the integration once it is set up, refer to [Using the Nextcloud integration](https://openproject.org/docs/user-guide/nextcloud-integration/).
 - For documentation on how to set up the integration as an administrator, refer to [Nextcloud integration setup](https://openproject.org/docs/system-admin-guide/integrations/nextcloud/).
 
+## Setting up the integration as an administrator with API
+We have a single API endpoint (/setup) available to set , update and delete the integration.
+1. Set up the whole integration with [POST] request
```
individual-it, line 26:
```suggestion
1. Set up the whole integration with a [POST] request
```

README.md
```diff
@@ -21,6 +21,68 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
...
+```json
+{
+    "nextcloud_oauth_client_name": <openproject-client>,
+    "openproject_redirect_uri": "http://<openproject_instance_url>/oauth_clients/<nextcloud_client_id>/callback",
+    "nextcloud_client_id": <nextcloud_client_id>,
+    "nextcloud_client_secret": <nextcloud_client_secret>
+}
+```
+
+2. Update the integration with [PUT] request
```
individual-it, line 49:
```suggestion
2. Update the integration with a [PUT] request
```

README.md
```diff
@@ -21,6 +21,68 @@ Please report issues and bugs here: https://community.openproject.org/projects/n
...
+```json
+{
+    "nextcloud_oauth_client_name": <openproject-client>,
+    "openproject_redirect_uri": "http://<openproject_instance_url>/oauth_clients/<nextcloud_client_id>/callback",
+    "nextcloud_client_id": <nextcloud_client_id>,
+    "nextcloud_client_secret": <nextcloud_client_secret>
+}
+```
+
+2. Resetting the whole integration with [DELETE] request
```
individual-it, line 72:
```suggestion
2. Resetting the whole integration with a [DELETE] request
```

individual-it (changes requested):

SwikritiT:
One of the things I noticed is that if we don't send the `Content-type : application/json ` header the request returns with `500` maybe `412` would be the appropriate error code.

lib/Controller/ConfigController.php
```diff
@@ -433,4 +445,106 @@ public function checkConfig(): DataResponse {
...
+				'error' => $e->getMessage()
+			], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
+
+
+	/**
+	 * @return array<mixed>
+	 */
+	public function recreateOauthClientInformation(): array {
```
individual-it, line 542: this is only used in the own class
```suggestion
	private function recreateOauthClientInformation(): array {
```

lib/Controller/ConfigController.php
```diff
@@ -145,14 +146,14 @@ public function setConfig(array $values): DataResponse {
 	 *
 	 * @param array<string, string|null> $values
 	 *
-	 * @return DataResponse
-	 * @throws OpenprojectErrorException
+	 * @return array<string, bool|string>
+	 * @throws \Exception
 	 */
-	public function setAdminConfig(array $values): DataResponse {
+	public function setUpIntegrationConfig(array $values): array {
```
individual-it, line 152: only used in the own class
```suggestion
	private function setIntegrationConfig(array $values): array {
```

lib/Service/OpenProjectAPIService.php
```diff
@@ -454,6 +455,56 @@ public static function validateURL(string $url): bool {
...
 
+	/**
+	 * validates the provided data for integration setup
+	 *
+	 * @param array<mixed> $values
+	 * @return bool
+	 * @throws \InvalidArgumentException
+	 *
+	 */
+	public static function validateIntegrationSetupInformation(array $values, ?string $keyType = null): bool {
```
individual-it, line 466: actually the second parameter is a bool, it only can have two values
```suggestion
	public static function validateIntegrationSetupInformation(array $values, bool $allKeysMandatory = true): bool {
```

individual-it (changes requested):

SwikritiT:
Tested manually. Works fine for me. No regression caught so far!

individual-it (approved):

PR #282, opened 2022-12-05
[OP##45055] stop showing OP notifications as NC notifications

individual-it:
As described in https://community.openproject.org/projects/nextcloud-integration/work_packages/45055 it does not really make sense to duplicate the notifications of OP to NC.
This PR removes all the support for showing NC notifications.

fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/45055

THE VIEW AND FUNCTIONALITY OF THE DASHBOARD AND THE NOTIFICATIONS THERE SHOULD NOT BE AFFECTED!


SwikritiT (approved):
LGTM

SwikritiT:
I think this can be removed as well https://github.com/nextcloud/integration_openproject/blob/32d3b8695da2d268cc8f5a88810a74e6f365710b/README.md?plain=1#L66

PR #285, opened 2022-12-06
Add Sabre plugin to whitelist the OP instance in the CORS headers

julien-nc:
Add a Sabre plugin that changes the DAV response headers on some conditions. This is how the https://github.com/digital-blueprint/webapppassword app does it.

:warning: We need https://github.com/nextcloud/server/pull/35621 for this PR to work with requests to `/public.php/webdav`.
Otherwise it just works with requests to `/remote.php/dav/files`.


individual-it:
I believe we can close this, as the direct upload was solved differently.
Or was there any other reason for this PR?

PR #287, opened 2022-12-12
[OP#42323] set pageSize for notifications to 'unlimited'

individual-it:
Without setting anything in the request URL the pageSize will be whatever is set in Administration=>System Settings=>General as	`Objects per page options`

Setting it to `-1` should solve the problem described in https://community.openproject.org/work_packages/42323/activity#activity-40


SwikritiT:
Screenshot from 2022-12-13 11-11-13
 looks good

SwikritiT (approved):

PR #290, opened 2022-12-14
[OP#45409] rewrite the reason strings to be the same as in OP

individual-it:
The strings for the reasons of notifications have been different in NC and OP because for most cases NC just showed what the API delivered. This PR changes the known strings and rewrites them to be the same as in OP

fixes https://community.openproject.org/notifications/details/45409


SagarGi (commented):
lgtm :+1:

PR #292, opened 2022-12-19
[OP#45465] tests for prepare direct upload API

individual-it:
This is a first set of API tests for https://community.openproject.org/projects/nextcloud-integration/work_packages/45465
They will not run in CI, as the API does not exist and also some test steps are not implemented yet, but those tests that can be already executed can be used for BDD to create the API

### DIFF
--- a/tests/acceptance/features/api/directUploadPrepare.feature
+++ b/tests/acceptance/features/api/directUploadPrepare.feature
@@ -3,18 +3,18 @@ Feature: API endpoint to prepare direct upload
 
   As an OpenProject user
   I want to upload files to Nextcloud from inside OpenProject
-  So that I get my stuff done faster
+  So that I don't need to leave OpenProject, open Nextcloud and then search for the same work package to create a link.
 
   As an OpenProject admin
-  I want the front-end to send the data to Nextcloud
-  So that the server load is minimized
+  I want the OpenProject front-end to send the files directly to Nextcloud
+  So that the long requests for uploading don't block any resources on the OpenProject back-end.
 
   Background:
     Given user "Alice" has been created
 
   Scenario Outline: Get a direct-upload token for a folder
     Given user "Alice" has created folder <folder>
-    When user "Alice" sends a GET request to the direct-upload endpoint with the id of <folder>
+    When user "Alice" sends a GET request to the direct-upload endpoint with the ID of <folder>
     Then the HTTP status code should be "200"
     And the data of the response should match
     """"
--- a/tests/acceptance/features/api/directUploadPrepare.feature
+++ b/tests/acceptance/features/api/directUploadPrepare.feature
@@ -34,7 +34,7 @@ Feature: API endpoint to prepare direct upload
       | folder                |
       | "/forOpenProject"     |
       | "/folder with spaces" |
-      | "/a/sub/folder        |
+      | "/a/sub/folder"       |
       | "/असजिलो folder"      |
       | "/?&$%?§ folder"      |
 
--- a/tests/acceptance/features/api/directUploadPrepare.feature
+++ b/tests/acceptance/features/api/directUploadPrepare.feature
@@ -40,7 +40,7 @@ Feature: API endpoint to prepare direct upload
 
 
   Scenario: Try to get a direct-upload token for the root folder
-    When user "Alice" sends a GET request to the direct-upload endpoint with the id of "/"
+    When user "Alice" sends a GET request to the direct-upload endpoint with the ID of "/"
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/api/directUploadPrepare.feature
```diff
@@ -0,0 +1,142 @@
+@skip
+Feature: API endpoint to prepare direct upload
+
+  As an OpenProject user
+  I want to upload files to Nextcloud from inside OpenProject
+  So that I get my stuff done faster
```
wielinde, line 6:
```suggestion
  So that I don't need to leave OpenProject, open Nextcloud and then search for the same work package to create a link..
```

tests/acceptance/features/api/directUploadPrepare.feature
```diff
@@ -0,0 +1,142 @@
...
+
+  As an OpenProject admin
+  I want the front-end to send the data to Nextcloud
```
wielinde, line 9:
```suggestion
  I want the OpenProject front-end to send the files directly to Nextcloud
```

tests/acceptance/features/api/directUploadPrepare.feature
```diff
@@ -0,0 +1,142 @@
...
+  So that the server load is minimized
```
wielinde, line 10:
```suggestion
  So that the long requests for uploading don't block any resources on the OpenProject back-end.
```

tests/acceptance/features/api/directUploadPrepare.feature
```diff
@@ -0,0 +1,142 @@
...
+
+  Background:
+    Given user "Alice" has been created
+
+  Scenario Outline: Get a direct-upload token for a folder
+    Given user "Alice" has created folder <folder>
```
wielinde, line 16:
```suggestion
    Given user "Alice" has created folder <folder> in Nextcloud
```

tests/acceptance/features/api/directUploadPrepare.feature
```diff
@@ -0,0 +1,142 @@
...
+    When user "Alice" sends a GET request to the direct-upload endpoint with the id of <folder>
```
wielinde, line 17:
```suggestion
    When user "Alice" sends a GET request to the direct-upload endpoint with the ID of <folder>
```

tests/acceptance/features/api/directUploadPrepare.feature
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+    """
+    Examples:
+      | folder                |
+      | "/forOpenProject"     |
+      | "/folder with spaces" |
+      | "/a/sub/folder        |
+      | "/असजिलो folder"      |
+      | "/?&$%?§ folder"      |
```
wielinde, line 39: I assume that the folder name and path is completely irrelevant. Is it necessary to cover the possible names for this test?

tests/acceptance/features/api/directUploadPrepare.feature
```diff
@@ -0,0 +1,142 @@
...
+
+
+  Scenario: Try to get a direct-upload token for the root folder
```
wielinde, line 42: Great to have it covered with a test!

wielinde (commented):
@individual-it The tests look good to me. Please find my tiny comments on wording.

wielinde:
Does the test miss that something is actually saved to the DB? Like the token, user_id, file_id, created_at, expires_on?

tests/acceptance/features/api/directUploadPrepare.feature
individual-it: I can make an other PR to adjust the wording, as this is a step that is already used in other tests

tests/acceptance/features/api/directUploadPrepare.feature
individual-it: They all should work, and as it does not cost much I'm suggesting to have those tests there just to make sure we don't have somewhere a bug that does weird things on UTF characters or special characters like `&`

individual-it:
> Does the test miss that something is actually saved to the DB? Like the token, user_id, file_id, created_at, expires_on?

From the point of view of the API we don't care what the system internally does with the data and how it saves it, we just care about the external behavior. In #293 we will test that the direct upload API actually does react correctly when we send it the token and use it for uploading data. So however it manages to achieve it from this test point its fine.

The implementation details can and should be tested in unit tests.

SwikritiT (approved):
LGTM

PR #294, opened 2022-12-20
move sharing related tests to new context

individual-it:
to make it easier to implement new tests and keep the FeatureContext shorter move the sharing related step to own SharingContext


SwikritiT (approved):
LGTM

SagarGi (commented):
looks good to me :+1:

PR #295, opened 2022-12-20
[OP#45399] make reason strings translatable

individual-it:
transifex did not pick up the new strings because they are not defined correctly, this PR should fix it


SwikritiT (approved):
LGTM

SagarGi (commented):
lgtm :+1:

PR #296, opened 2022-12-20
[OP#45465] Feature/prepare direct upload api

SwikritiT:
Related work-package: [OP#45465] https://community.openproject.org/projects/nextcloud-integration/work_packages/45465/activity?query_id=3504

## Description
This PR adds a new endpoint `/direct-upload` that tests the parameter `folder_id`, for the preparation of the direct upload.
The added endpoint will return the `token` that can be used for direct upload and `expires_on` which is the expiration time of the token.

The end-point can be tested with:

```console
 curl -u USER:PASSWD http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload?folder_id=<folder_id>
```

### DIFF
--- a/lib/Controller/DirectUploadController.php
+++ b/lib/Controller/DirectUploadController.php
@@ -24,7 +24,6 @@
 
 namespace OCA\OpenProject\Controller;
 
-
 use OC\User\NoUserException;
 use OCA\OpenProject\Service\DirectUploadService;
 use OCP\AppFramework\Controller;
@@ -34,14 +33,12 @@ use OCP\DB\Exception;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 use OCP\Files\NotPermittedException;
-use OCP\IL10N;
 use OCP\IRequest;
 use OCP\IUser;
 use OCP\IUserSession;
-use Psr\Log\LoggerInterface;
 use OCP\Files\FileInfo;
 
-class DirectUploadController extends Controller{
+class DirectUploadController extends Controller {
 	/**
 	 * @var string|null
 	 */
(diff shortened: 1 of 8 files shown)


lib/Controller/DirectUploadController.php
```diff
@@ -0,0 +1,114 @@
...
+	/**
+	 * @var IRootFolder
+	 */
+	private $rootFolder;
+	public function __construct(string $appName,
+								IRequest $request,
+								IRootFolder $rootFolder,
+								IUserSession $userSession,
+								DirectUploadService $directUploadService,
+								?string $userId) {
```
individual-it, line 66: indention is strange

lib/Controller/DirectUploadController.php
```diff
@@ -0,0 +1,114 @@
...
+				$fileType === FileInfo::TYPE_FOLDER
+			) {
+				$response = $this->directUploadService->getTokenForDirectUpload($folder_id, $this->userId);
+				return new DataResponse($response);
+			} else {
+				return new DataResponse([
+					'error' => 'folder not found or not enough permissions'
+				], Http::STATUS_NOT_FOUND);
+			}
+		} catch (Exception|NoUserException|NotPermittedException|NotFoundException $e) {
```
individual-it, line 108: if you catch `Exception` there is no need to list the other ones as they all extend `Exception`
```suggestion
		} catch (Exception $e) {
```

lib/Service/DirectUploadService.php
```diff
@@ -0,0 +1,99 @@
...
+
+	/** @var string table name */
+	private $table = 'directUpload';
+
+	/** @var string time of token expiration */
+	private $expiryTime = '+1 hour';
+
+	public function __construct(IDBConnection $db,
+								IL10N $l,
+								ISecureRandom $secureRandom) {
```
individual-it, line 57: also weird indent

PR #303, opened 2022-12-22
[OP#45464]Direct upload implementation

SwikritiT:
[OP#45464] : https://community.openproject.org/projects/nextcloud-integration/work_packages/45464

This PR is the second part of the direct-upload feature.
After getting the token from the `preparation` endpoint send multipart form data POST request to this endpoint to upload the file

## Using Curl
```bash
curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
--form 'file=@"<path-of-file>"' -H'Content-Type: multipart/form-data'
```


SwikritiT:
Opening this up for review. Meanwhile, I'll add more tests to satisfy the PHP coverage

tests/acceptance/features/api/directUpload.feature
```diff
@@ -33,12 +32,12 @@ Feature: API endpoint for direct upload
...
+    And the content of file at "<file-name>" for user "Alice" should be "some data"
     Examples:
-      | file-name         |
-      | "textfile0.txt"   |
-      | "असजिलो file"     |
-      | "?&$%?§ file.txt" |
+      | file-name     |
+      | textfile0.txt |
+      | असजिलो file   |
+#      | ?&$%?§ file.txt |
```
individual-it, line 40: what is wrong with that file-name?

lib/Controller/DirectUploadController.php
```diff
@@ -104,10 +126,107 @@ public function prepareDirectUpload(int $folder_id):DataResponse {
...
+	 * @param string $contents
+	 *
+	 * @return DataResponse
+	 */
+	public function directUpload(string $token, string $file_name, string $contents):DataResponse {
+		$fileId = null;
+		try {
+			if (strlen($token) !== 64 || !preg_match('/^[a-zA-Z0-9]*/', $token)) {
+				return new DataResponse([
+					'error' => 'Invalid token. The token should be 64 characters long and in human readable format.'
```
individual-it, line 158: not sure if we want to tell anyone who is guessing tokens what the format should be. Use the token that you generated, if you send anything else you will receive a `'Invalid token'` message.
That check could also be integrated into `getTokenInfo()` or you could throw an exception here `throw new NotFoundException('Invalid token.');` then there will be just one place where the DataResponse is generated

lib/Service/DatabaseService.php
```diff
@@ -0,0 +1,119 @@
...
+		$query = $this->db->getQueryBuilder();
+		$query->select('user_id', 'created_at', 'expires_on', 'folder_id')
+			->from($this->table)
+			->where(
+				$query->expr()->eq('token', $query->createNamedParameter($token))
+			);
+		$req = $query->executeQuery();
+		while ($row = $req->fetch()) {
+			$userId = $row['user_id'];
+			$expiration = (int) $row['expires_on'];
```
individual-it, line 92: just to reduce brain work, the logical naming would be:
```suggestion
			$expiresOn = (int) $row['expires_on'];
```

tests/acceptance/features/bootstrap/DirectUploadContext.php
```diff
@@ -69,16 +69,23 @@ public function userGotADirectUploadTokenFor(
 	public function anonymousUserSendsAMultipartFormDataPOSTRequestToTheEndpointWith(
 		string $endpoint, TableNode $formData
 	): void {
-		$endpoint = $this->replaceInlineCodes($endpoint);
+		if (str_contains($endpoint, '%last-created-direct-upload-token%')) {
+			$endpoint = $this->replaceInlineCodes($endpoint);
+		}
```
individual-it, line 74: why not sending it all to the `replaceInlineCodes` ? If there is nothing to replace it would come back as it is?

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -729,14 +728,14 @@ public function sendRequestsToAppEndpoint(
 	) {
 		$fullUrl = $this->getBaseUrl();
 		$fullUrl .= "index.php/apps/integration_openproject/" . $endpoint;
-		if ($headers === null) {
-			$headers['Accept'] = 'application/json';
-		}
+//		if ($headers === null) {
+//			$headers['Accept'] = 'application/json';
+//		}
```
individual-it, line 733: I think we need that header for other tests

tests/lib/Controller/DirectUploadControllerTest.php
```diff
@@ -0,0 +1,119 @@
...
+			],
+			$result->getData()
+		);
+		assertSame(404, $result->getStatus());
+	}
+
+//	public function testDirectUpload():void{
+//
+//	}
+	/**
```
individual-it, line 53: delete code that is commented out

tests/lib/Controller/DirectUploadControllerTest.php
```diff
@@ -0,0 +1,119 @@
...
+		$directUploadServiceMock->method('getTokenForDirectUpload')
+			->willReturn([
+				'token' => 'WampxL5Z97CndGwB7qLPfotosDT5mXk7oFyGLa64nmY35ANtkzT7zDQwYyXrbdC3',
+				'expires_on' => 1671537939
+			]);
+//		$directUploadServiceMock->method('getTokenInfo')->willReturn(
+//			[
+//
+//			]
+//		);
```
individual-it, line 82: delete if not needed

individual-it (changes requested):

lib/Service/DirectUploadService.php
```diff
@@ -46,55 +44,77 @@ class DirectUploadService {
...
+	 *
+	 * @return array<mixed>
+	 *
+	 * @throws NotPermittedException
+	 * @throws NotFoundException
+	 * @throws Exception
+	 */
+	public function getTokenInfo(string $token): ?array {
+		$tokenInfo = $this->databaseService->getTokenInfoFromDB($token);
+		if ($tokenInfo['user_id'] === null || !$this->userManager->userExists($tokenInfo['user_id'])) {
```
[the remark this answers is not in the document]
SwikritiT: Tested this manually and the token from the disabled user works. I'll add a check so that it doesn't happen and add a test as well

lib/Controller/DirectUploadController.php
SwikritiT: I didn't think it that way. I'll refactor it

tests/acceptance/features/bootstrap/DirectUploadContext.php
SwikritiT: Yes but we are replacing the string with the last created upload token https://github.com/nextcloud/integration_openproject/blob/ab9f8a4c83f0fac355c7849faa4219f9512e4882/tests/acceptance/features/bootstrap/DirectUploadContext.php#L122
and when testing for invalid token  (hard-coded -token-from-feature-file) this ^ function throws error

tests/acceptance/features/bootstrap/FeatureContext.php
SwikritiT: If I enable this the test will never enter the next if statement 
https://github.com/nextcloud/integration_openproject/blob/ab9f8a4c83f0fac355c7849faa4219f9512e4882/tests/acceptance/features/bootstrap/FeatureContext.php#L738

Because by the time it reaches here ^ one of the conditions i.e `$headers===null` is already false because `$headers` is set by the above commented if block.

the headers are set in else block. Do we really need that if check?

tests/acceptance/features/api/directUpload.feature
```diff
@@ -16,9 +15,37 @@ Feature: API endpoint for direct upload
...
+    """
+    And the content of file at "<file-name>" for user "Alice" should be "some data"
+    Examples:
+      | file-name       | pattern                     |
+      | textfile0.txt   | textfile0\\.txt             |
+      | असजिलो file     | असजिलो file                 |
+      | ?&$%?§ file.txt | \\?\\&\\$\\%\\?§ file\\.txt |
+
+
+  Scenario Outline: Send valid file-names to the direct-upload endpoint
```
individual-it, line 43: I think we could just integrate this scenario into the one above

tests/acceptance/features/api/directUpload.feature
```diff
@@ -208,13 +231,56 @@ Feature: API endpoint for direct upload
...
+        "error"
+      ],
+      "properties": {
+          "error": {"type": "string", "pattern": "^unauthorized$"}
+      }
+    }
+    """
+
 
+  Scenario: use the token of disable user
```
individual-it, line 263:
```suggestion
  Scenario: use a token created by a user that was disabled after creating the token
```

individual-it: we should also check a deleted user

lib/Controller/DirectUploadController.php
```diff
@@ -91,10 +114,11 @@ public function prepareDirectUpload(int $folder_id):DataResponse {
...
 				], Http::STATUS_NOT_FOUND);
 			}
-			$node = array_shift($nodes);
-			$fileType = $node->getType();
+			$folderNode = array_shift($nodes);  // @phpstan-ignore-line
+			$fileType = $folderNode->getType();  // @phpstan-ignore-line
+			// @phpstan-ignore-next-line
 			if (
-				$node->isCreatable() &&
+				$folderNode->isCreatable() &&
```
individual-it, line 121: all the changes in this function do not really affect the new functionality, should we revert them?

lib/Controller/DirectUploadController.php
```diff
@@ -104,10 +128,100 @@ public function prepareDirectUpload(int $folder_id):DataResponse {
 					'error' => 'folder not found or not enough permissions'
 				], Http::STATUS_NOT_FOUND);
 			}
-		} catch (Exception $e) {
+		} catch (Exception | NotPermittedException |NoUserException $e) {
```
individual-it, line 131: catching "Exception" will also catch all other possible exceptions

lib/Controller/DirectUploadController.php
```diff
@@ -104,10 +128,100 @@ public function prepareDirectUpload(int $folder_id):DataResponse {
...
+			$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+			$nodes = $userFolder->getById($tokenInfo['folder_id']);
+			if (empty($nodes)) {
+				return new DataResponse([
+					'error' => 'folder not found or not enough permissions'
+				], Http::STATUS_NOT_FOUND);
+			}
+			$folderNode = array_shift($nodes);  // @phpstan-ignore-line
+			if (
+				$folderNode->isCreatable()  // @phpstan-ignore-line
```
individual-it, line 171: phpstan does not complain for me when deleting the `// @phpstan-ignore-line` comments

tests/acceptance/features/api/directUpload.feature
```diff
@@ -102,7 +125,7 @@ Feature: API endpoint for direct upload
         "error"
       ],
       "properties": {
-          "error": {"type": "string", "pattern": "^conflict$"}
+          "error": {"type": "string", "pattern": "^Conflict, file with name file.txt already exists.$"}
```
individual-it, line 128: I wouldn't mention the file name, the user knows it

tests/acceptance/features/bootstrap/FeatureContext.php
individual-it: you are right, let's delete that whole if check

tests/acceptance/features/bootstrap/DirectUploadContext.php
individual-it: Oh yes, `str_replace` calls `getLastCreatedDirectUploadToken()` even nothing is found. 
Let's move the check into `replaceInlineCodes()` and I can work on refactoring the `replaceInlineCodes()` to get the other codes like `%now+3500s%` and  `%ids[1]%` into it

lib/Service/DirectUploadService.php
```diff
@@ -46,55 +44,77 @@ class DirectUploadService {
...
+	 */
+	public function getTokenInfo(string $token): ?array {
+		$tokenInfo = $this->databaseService->getTokenInfoFromDB($token);
+		$userId = $this->userManager->get($tokenInfo['user_id']);
+		if ($tokenInfo['user_id'] === null || !$this->userManager->userExists($tokenInfo['user_id']) || !$userId->isEnabled()) {
+			throw new NotPermittedException('unauthorized');
+		}
+		$currentTime = (new DateTime())->getTimestamp();
+		if ($currentTime > $tokenInfo['expires_on']) {
+			throw new NotFoundException('Invalid token.');
```
individual-it, line 116: we should unify the error messages, some are starting with a capital letter, some not, some have a `.` at the end, some not

individual-it (changes requested):
some small nitty picky comments

tests/acceptance/features/api/directUpload.feature
SwikritiT: test added in new commit

lib/Controller/DirectUploadController.php
SwikritiT: true. I've reverted them

lib/Controller/DirectUploadController.php
SwikritiT: the CI complains

tests/acceptance/features/api/directUpload.feature
SwikritiT: removed

lib/Service/DirectUploadService.php
SwikritiT: true. I've made them similar

individual-it (approved):

PR #306, opened 2022-12-29
[OP#45551] use `shivammathur/setup-php@2.23.0` instead of `shivammathur/setup-php@v2`

SwikritiT:
We were having a `Segmentation fault  (core dumped)` error in CI upon using the `2.23.0` tag of `setup-php` instead of `v2` the problem doesn't occur. `2.23.0` is the latest version now. 

Related work package : [OP#45551] https://community.openproject.org/projects/nextcloud-integration/work_packages/45551


SagarGi (commented):
LGTM :+1:

SagarGi (approved):

PR #307, opened 2023-01-04
mention all notification reasons

individual-it:
All possible strings need to be mentioned so that transifex picks them up correctly and translates them.

This should finally fix https://community.openproject.org/projects/nextcloud-integration/work_packages/45399


SagarGi (commented):
LGTM :+1:

SagarGi (approved):

PR #308, opened 2023-01-04
move all replacing of inline codes into one place

individual-it:
Move all replacements of inline codes of API tests to one function


SwikritiT (approved):

SagarGi (commented):
code LGTM :+1:

PR #310, opened 2023-01-05
[OP#45464] Feature/direct upload api overwrite

SwikritiT:
related work package [OP#45464] : https://community.openproject.org/projects/nextcloud-integration/work_packages/45464

## Description
**Direct upload with overwrite parameter**:
   The overwrite parameter can be either set to `true` or `false`.
      1. **overwrite set to false**: 
      If the parameter is set to `false` and a file with the name already exists, a new file will be uploaded with existing name having a number suffix.
```console
curl-X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
--form 'file=@"<path-of-file>"' --form 'overwrite="false"' -H'Content-Type: multipart/form-data' 
```
The response from the above curl request will be
```json
 {
  "file_name": "<file_name>(some-number)",
   "file_id": <file_id>
 }  
```
     
 2. **overwrite set to true**:
      If the parameter is set to `true` and a file with the name already exists, the existing file will be overwritten.
 ```console
 curl-X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
 --form 'file=@"<path-of-file>"' --form 'overwrite="true"' -H'Content-Type: multipart/form-data'
 ```

The response from the above curl request will be
```json
{
   "file_name": "<file_name>",
  "file_id": <file_id>
} 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
       ```
    
    2. **Direct upload with overwrite parameter**:
-   The overwrite parameter can be either set to `true` or `false`.
+      The overwrite parameter can be either set to `true` or `false`.
       1. **overwrite set to false**: 
          If the parameter is set to `false` and a file with the name already exists, a new file will be uploaded with existing name having a number suffix.
 	      ```console
@@ -170,6 +170,7 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
 	      ```
       2. **overwrite set to true**:
          If the parameter is set to `true` and a file with the name already exists, the existing file will be overwritten.
+
 		   ```console
 		   curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
 		   --form 'file=@"<path-of-file>"' --form 'overwrite="true"' -H'Content-Type: multipart/form-data'
--- a/README.md
+++ b/README.md
@@ -169,12 +169,12 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
 	          "file_id": <file_id>
 	      }  
 	      ```
-      2. **overwrite set to true**:
-         If the parameter is set to `true` and a file with the name already exists, the existing file will be overwritten.
 
-		   ```console
-		   curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
-		   --form 'file=@"<path-of-file>"' --form 'overwrite="true"' -H'Content-Type: multipart/form-data'
+      2. **overwrite set to true**: 
+         If the parameter is set to `true` and a file with the name already exists, the existing file will be overwritten.
+	      ```console
+         curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
+         --form 'file=@"<path-of-file>"' --form 'overwrite="true"' -H'Content-Type: multipart/form-data' 
 		   ```
 
 	      The response from the above curl request will be
--- a/README.md
+++ b/README.md
@@ -138,48 +138,48 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
 2. **Direct upload**:
    Send multipart form data POST request to `direct-upload` end-point to upload the file with `token` acquired from preparation endpoint. The API takes and optional parameter `overwrite`.
    1. **Direct upload without overwrite parameter**:
-   If the `overwrite` parameter is not set at all only the file with non-existing name is uploaded and a conflict error is thrown in case the file already exists.
-         ```console
-          curl-X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
-          --form 'file=@"<path-of-file>"' -H'Content-Type: multipart/form-data'
-        ```
-      
-        The response from the above curl request will be
-        ```json
-        {
+      In case the `overwrite` parameter is not set at all a file will be only uploaded if no file exists with that name, otherwise a conflict error is thrown.
+      ```console
+      curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
+      --form 'file=@"<path-of-file>"' -H'Content-Type: multipart/form-data'
+      ```
+
+      The response from the above curl request will be
+      ```json
+      {
           "file_name": "<file_name>",
           "file_id": <file_id>
-        }  
-       ```
+      }  
+      ```
    
    2. **Direct upload with overwrite parameter**:
    The overwrite parameter can be either set to `true` or `false`.
       1. **overwrite set to false**: 
-      If the parameter is set to `false` and a file with the name already exists, a new file will be uploaded with existing name having a number suffix.
-	     ```console
-          curl-X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
-          --form 'file=@"<path-of-file>"' --form 'overwrite="false"' -H'Content-Type: multipart/form-data' 
-		 ```
+         If the parameter is set to `false` and a file with the name already exists, a new file will be uploaded with existing name having a number suffix.
+	      ```console
+         curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
+         --form 'file=@"<path-of-file>"' --form 'overwrite="false"' -H'Content-Type: multipart/form-data' 
+		   ```
 
 	      The response from the above curl request will be
 	      ```json
 	      {
-	      "file_name": "<file_name>(some-number)",
-	      "file_id": <file_id>
+	          "file_name": "<file_name>(some-number)",
+	          "file_id": <file_id>
 	      }  
 	      ```
       2. **overwrite set to true**:
-      If the parameter is set to `true` and a file with the name already exists, the existing file will be overwritten.
-		 ```console
-		 curl-X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
-		 --form 'file=@"<path-of-file>"' --form 'overwrite="true"' -H'Content-Type: multipart/form-data'
-		 ```
+         If the parameter is set to `true` and a file with the name already exists, the existing file will be overwritten.
+		   ```console
+		   curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
+		   --form 'file=@"<path-of-file>"' --form 'overwrite="true"' -H'Content-Type: multipart/form-data'
+		   ```
 
 	      The response from the above curl request will be
 	      ```json
 	      {
-	      "file_name": "<file_name>",
-	      "file_id": <file_id>
+	          "file_name": "<file_name>",
+	          "file_id": <file_id>
 	      }  
 	      ```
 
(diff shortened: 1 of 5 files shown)


lib/Controller/DirectUploadController.php
```diff
@@ -165,13 +171,35 @@ public function directUpload(string $token):DataResponse {
...
+					$file = $folderNode->get($fileName); // @phpstan-ignore-line
+					if (!$file->isUpdateable()) {
+						return new DataResponse([
+							'error' => "not enough permissions"
+						], Http::STATUS_FORBIDDEN);
+					}
+					// overwrite the file
+					$file->putContent(fopen($tmpPath, 'r'));
+					$fileId = $file->getId();
+					$this->databaseService->deleteToken($token);
```
individual-it, line 184: what should we do in the case of an error, still delete the token, or leave it as it is?

tests/acceptance/features/api/directUpload.feature
```diff
@@ -351,15 +351,15 @@ Feature: API endpoint for direct upload
         "file_id"
       ],
       "properties": {
-          "file_name": {"type": "string", "pattern": "^file \\(2\\)\\(2\\)\\.txt$"},
+          "file_name": {"type": "string", "pattern": "^file \\(3\\)\\.txt$"},
```
individual-it, line 354: I guess that is the choice `getNonExistingName()` does, so its the standard way for NC

lib/Controller/DirectUploadController.php
SwikritiT: In the specification, the deletion of the token is only after the successful upload, I think we can leave it as it is for errors so that the token can be reused for the next request.

README.md
```diff
@@ -135,19 +135,53 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
-       "file_name": "<file_name>",
-       "file_id": <file_id>
-   }  
-   ```
+2. **Direct upload**:
+   Send multipart form data POST request to `direct-upload` end-point to upload the file with `token` acquired from preparation endpoint. The API takes and optional parameter `overwrite`.
+   1. **Direct upload without overwrite parameter**:
+   If the `overwrite` parameter is not set at all only the file with non-existing name is uploaded and a conflict error is thrown in case the file already exists.
+         ```console
+          curl-X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
```
individual-it, line 143: there needs to be a space after `curl`, I've made a similar fix in [09a0e47](https://github.com/nextcloud/integration_openproject/commit/09a0e4749df929ed7a41450326ade49af6582ad2)
```suggestion
          curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
```
same in the other curl examples

README.md
```diff
@@ -135,19 +135,53 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
+
+	      The response from the above curl request will be
+	      ```json
+	      {
+	      "file_name": "<file_name>(some-number)",
+	      "file_id": <file_id>
+	      }  
+	      ```
+      2. **overwrite set to true**:
+      If the parameter is set to `true` and a file with the name already exists, the existing file will be overwritten.
```
individual-it, line 172: What happens if there is a folder with that name?

README.md
```diff
@@ -135,19 +135,53 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
+		 ```console
```
individual-it, line 173: there seems to be a mixture of spaces and tabs for indenting in your change, I think its better to use spaces everywhere

lib/Controller/DirectUploadController.php
```diff
@@ -165,13 +178,33 @@ public function directUpload(string $token):DataResponse {
...
 			) {
 				// @phpstan-ignore-next-line
-				if ($folderNode->nodeExists($fileName)) {
+				if ($folderNode->nodeExists($fileName) && $overwrite) {
+					$file = $folderNode->get($fileName); // @phpstan-ignore-line
+					if (!$file->isUpdateable()) {
+						throw new ForbiddenException('not enough permissions');
+					}
+					// overwrite the file
+					$file->putContent(fopen($tmpPath, 'r'));
```
individual-it, line 187: do we need to delete the `$tmpPath` afterwards?

lib/Controller/DirectUploadController.php
```diff
@@ -147,6 +149,17 @@ public function directUpload(string $token):DataResponse {
 		try {
 			$fileId = null;
 			$directUploadFile = $this->request->getUploadedFile('file');
+			$acceptedOverwriteValues = ['true','True','false','False'];
+			$overwrite = $this->request->getParam('overwrite');
+			if (isset($overwrite)) {
+				if (in_array($overwrite, $acceptedOverwriteValues)) {
+					$overwrite = $overwrite === 'true' || $overwrite === 'True' ;
```
individual-it, line 156: what about running `$overwrite = strtolower($overwrite);` then we don't need to check all kind of possibilities

individual-it (changes requested):

README.md
SwikritiT: I don't know, I would expect the file to be uploaded. I'll try it and add tests as well.

SwikritiT: The current code throws `500` when overwrite is set to `true`, and works fine when set to `false` or when not set. I'll fix the code so it doesn't send `500` in this case

lib/Controller/DirectUploadController.php
SwikritiT: Maybe not according to this documentation https://www.php.net/manual/en/features.file-upload.post-method.php
> The file will be deleted from the temporary directory at the end of the request if it has not been moved away or renamed.

tests/acceptance/features/api/directUpload.feature
```diff
@@ -458,3 +458,105 @@ Feature: API endpoint for direct upload
...
+      "properties": {
+          "file_name": {"type": "string", "pattern": "^file \\(2\\)\\.txt$"},
+          "file_id": {"type" : "integer"}
+      }
+    }
+    """
+    And the content of file at "/file (2).txt" for user "Alice" should be "new data"
+
+
+  Scenario: don't set override and send a file with existing folder name
```
SwikritiT, line 544:
```suggestion
  Scenario: don't set overwriteand send a file with an existing folder name
```

README.md
```diff
@@ -134,20 +134,56 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
        "expires_on": <some_timestamp>
    }
    ```
+   > Note: The token is one-time only.
+
+2. **Direct upload**:
+   Send multipart form data POST request to `direct-upload` end-point to upload the file with `token` acquired from preparation endpoint. The API takes and optional parameter `overwrite`.
```
phil-davis, line 140:
```suggestion
   Send a multipart form data POST request to the `direct-upload` end-point to upload the file with `token` acquired from the preparation endpoint. The API takes an optional parameter `overwrite`.
```

README.md
```diff
@@ -134,20 +134,56 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
+   1. **Direct upload without overwrite parameter**:
+      In case the `overwrite` parameter is not set at all a file will be only uploaded if no file exists with that name, otherwise a conflict error is thrown.
```
phil-davis, line 142:
```suggestion
      If the `overwrite` parameter is not set, a file will only be uploaded if no file exists with that name, otherwise a conflict error is thrown.
```

README.md
```diff
@@ -134,20 +134,56 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
+      {
+          "file_name": "<file_name>",
+          "file_id": <file_id>
+      }  
+      ```
+   
+   2. **Direct upload with overwrite parameter**:
+      The overwrite parameter can be either set to `true` or `false`.
+      1. **overwrite set to false**: 
+         If the parameter is set to `false` and a file with the name already exists, a new file will be uploaded with existing name having a number suffix.
```
phil-davis, line 159:
```suggestion
         If the parameter is set to `false` and a file with the name already exists, a new file will be uploaded with the existing name and a number suffix.
```

README.md
```diff
@@ -122,7 +122,7 @@ bash integration_setup.sh
 ## Direct upload
 There's an end-point `direct-upload` available which can be used for direct-upload. There's two steps to direct upload first we need to get the `token` which is then used for direct upload.
```
phil-davis, line 123:
```suggestion
There's an end-point `direct-upload` available which can be used for direct-upload. There's two steps to direct upload. first we need to get the `token`. Then use the token in the direct upload request.
```

README.md
```diff
@@ -134,20 +134,56 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
+         If the parameter is set to `false` and a file with the name already exists, a new file will be uploaded with existing name having a number suffix.
+          ```console
+         curl -X POST 'http://<nextcloud_host>/index.php/apps/integration_openproject/direct-upload/<token>' \
+         --form 'file=@"<path-of-file>"' --form 'overwrite="false"' -H'Content-Type: multipart/form-data' 
+           ```
+
+          The response from the above curl request will be
+          ```json
+          {
+              "file_name": "<file_name>(some-number)",
```
phil-davis, line 168: What happens about "file extension". Do we get `file(42).doc` or `file.doc(42)`?

Maybe there could be a real example that shows this.

README.md
```diff
@@ -134,20 +134,64 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
+           ```
+
+          The response from the above curl request will be
+          ```json
+          {
+              "file_name": "<file_name>(some-number).ext",
+              "file_id": <file_id>
+          }
+          ```
+          Suppose we have a file with name `file.txt` in the server, and we send a request to direct-upload api with file `file.txt` and overwrite set to `true` then the response will be.
```
individual-it, line 172:
```suggestion
          Suppose we have a file with name `file.txt` in the server, and we send a request to direct-upload api with the file `file.txt` and overwrite is set to `false` then the response will be.
```

README.md
```diff
@@ -134,20 +134,64 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
+		  ```json
+		  {
+			  "file_name": "file (2).txt",
+			  "file_id": 123
+		  }
+		  ```
+          > Note: The file id in this case is the id of the original file. Only the content is overwritten.
```
individual-it, line 179: I think this line has to go to `**overwrite set to true**:`

individual-it (changes requested):

README.md
SwikritiT: yes sorry I didn't notice

individual-it (approved):

README.md
```diff
@@ -135,19 +135,53 @@ There's an end-point `direct-upload` available which can be used for direct-uplo
...
-   ```json
-   {
-       "file_name": "<file_name>",
-       "file_id": <file_id>
-   }  
-   ```
+2. **Direct upload**:
+   Send multipart form data POST request to `direct-upload` end-point to upload the file with `token` acquired from preparation endpoint. The API takes and optional parameter `overwrite`.
+   1. **Direct upload without overwrite parameter**:
+   If the `overwrite` parameter is not set at all only the file with non-existing name is uploaded and a conflict error is thrown in case the file already exists.
```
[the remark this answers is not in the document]
phil-davis: We now have:
"If the `overwrite` parameter is not set, a file will only be uploaded if no file exists with that name, otherwise a conflict error is thrown."

That looks good to me.

PR #311, opened 2023-01-05
[OP#45464] API tests for invalid overwrite values

individual-it:
Some more tests for cases where the "overwrite" values is invalid

### DIFF
--- a/tests/acceptance/features/api/directUpload.feature
+++ b/tests/acceptance/features/api/directUpload.feature
@@ -410,6 +410,42 @@ Feature: API endpoint for direct upload
     And the content of file at "/file.txt" for user "Alice" should be "original data"
 
 
+  @skip
+  Scenario Outline: set overwrite to an invalid value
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/api/directUpload.feature
```diff
@@ -410,6 +410,42 @@ Feature: API endpoint for direct upload
...
+          "error": {"type": "string", "pattern": "^invalid overwrite value$"}
+      }
+    }
+    """
+    And the content of file at "/file.txt" for user "Alice" should be "original data"
+    Examples:
+      | overwrite |
+      | 1         |
+      | 0         |
+      | null      |
```
SwikritiT, line 445: if the overwrite is null should it be the normal case like "without over-write parameter"?

individual-it: this will send a string containing "null", so its an invalid string
also the next test sends an empty string, what should not be accepted

SwikritiT (approved):

PR #313, opened 2023-01-06
[OP#45596] fix fetching avatars from openproject

individual-it:
This PR fixes the fetching of avatars from openproject.
The problem was that we had the wrong userId in `getOpenProjectAvatar`, the id given was the openproject id, but we tried to use it to read the token from the NC database.
Now we passing also in the NC userId to read the token and the openproject userId is used to generate the openproject URL to get the avatars from there as before.


SagarGi (commented):
code LGTM :+1:

SagarGi (approved):

PR #323, opened 2023-01-18
seperate API tests and run them on different DB

individual-it:
The bug that made #320 necessary was not discovered by the tests, because the tests never run with pqsql.
This PR runs the API tests in a separate job and uses real (mysql,pgsql) databases for that.

~~currently blocked by https://github.com/juliushaertl/nextcloud-docker-dev/issues/142~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -249,6 +249,5 @@ jobs:
           NEXTCLOUD_BASE_URL: http://nextcloud
         run: |
           composer install --no-progress --prefer-dist --optimize-autoloader
-          counter=0; until [ $(curl -s -f http://nextcloud/status.php | grep '"installed":true') ] || [ $counter -gt 12 ] ; do echo .; sleep 10; counter=$((counter+1)); done
-          if [ $counter -gt 12 ]; then exit 1; fi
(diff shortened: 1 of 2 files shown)


juliushaertl:
New images have been pushed in https://github.com/juliushaertl/nextcloud-docker-dev/actions/runs/4003588199/jobs/6871824562

.github/workflows/ci.yml
```diff
@@ -158,3 +144,110 @@ jobs:
...
+            phpVersionMajor: 8
+            phpVersionMinor: 1
+          - nextcloudVersion: master
+            phpVersionMajor: 7
+          - phpVersionMajor: 7
+            phpVersionMinor: 0
+          - phpVersionMajor: 7
+            phpVersionMinor: 1
+          - phpVersionMajor: 8
+            phpVersionMinor: 4
```
SwikritiT, line 172: what do thye do?

SwikritiT, line 172: what do they do?

individual-it: the matrix will create all possible combinations, but we don't want to test PHP 7.0, 7.1, 8.4

SwikritiT (approved):

PR #325, opened 2023-01-19
[OP#45849] check quota before writing file

individual-it:
Check for free quota on that particular folder before writing a file.
There is still the chance of a race-condition. e.g. while we are writing data an other client also writes and together we run out of quota, no idea how we could avoid that.

part of https://community.openproject.org/projects/nextcloud-integration/work_packages/45849

### DIFF
--- a/lib/Controller/DirectUploadController.php
+++ b/lib/Controller/DirectUploadController.php
@@ -24,12 +24,15 @@
 
 namespace OCA\OpenProject\Controller;
 
+use OC\Files\Node\Folder;
 use OC\User\NoUserException;
 use InvalidArgumentException;
 use OC\ForbiddenException;
 use \OCP\AppFramework\ApiController;
+use OCP\Files\File;
 use OCP\Files\InvalidCharacterInPathException;
 use OCP\Files\InvalidPathException;
+use OCP\Files\NotEnoughSpaceException;
 use OCP\Files\NotFoundException;
 use OCA\OpenProject\Service\DirectUploadService;
 use OCP\AppFramework\Http;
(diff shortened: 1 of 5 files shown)


lib/Controller/DirectUploadController.php
```diff
@@ -142,6 +145,7 @@ public function directUpload(string $token):DataResponse {
 		try {
 			$fileId = null;
 			$directUploadFile = $this->request->getUploadedFile('file');
+			$tmpPath = $directUploadFile['tmp_name'];
```
SwikritiT, line 148: Maybe `$filename` can also be brought up here.

lib/Controller/DirectUploadController.php
```diff
@@ -167,12 +171,26 @@ public function directUpload(string $token):DataResponse {
...
 			if (
 				$folderNode->isCreatable()
 			) {
 				// @phpstan-ignore-next-line
 				if ($folderNode->nodeExists($fileName) && $overwrite) {
+					/**
+					 * @var File $file
+					 */
 					$file = $folderNode->get($fileName); // @phpstan-ignore-line
 					if ($file->getType() === FileInfo::TYPE_FOLDER) {
```
SwikritiT, line 195: Maybe these can be removed now?
```suggestion
					$file = $folderNode->get($fileName);
					if ($file->getType() === FileInfo::TYPE_FOLDER) {
```

SagarGi (commented):
looks good to me except the above comments.

lib/Controller/DirectUploadController.php
individual-it: after #328 I will make a separate PR that will hopefully get rid of all the `@phpstan-ignore-line` bits

lib/Controller/DirectUploadController.php
individual-it: done

SwikritiT (approved):

PR #328, opened 2023-01-20
[OP#45849] report error when file was not uploaded

individual-it:
~~on top of #325~~
This case happens when the uploaded file is bigger than upload_max_filesize.
In that case return a 413 HTTP code and show what the max upload size is.

### DIFF
--- a/tests/lib/Controller/DirectUploadControllerTest.php
+++ b/tests/lib/Controller/DirectUploadControllerTest.php
@@ -142,7 +142,7 @@ class DirectUploadControllerTest extends TestCase {
 		$userFolderMock = $this->getMockBuilder('\OCP\Files\Folder')->getMock();
 		$userFolderMock->method('getById')->willReturn($nodeMock);
 		$directUploadController = $this->createDirectUploadController(
-			$userFolderMock, 100, ''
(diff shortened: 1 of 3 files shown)


tests/lib/Controller/DirectUploadControllerTest.php
```diff
@@ -120,12 +120,51 @@ public function testDirectUploadNotEnoughSpace():void {
...
+	/**
+	 * @return array<int, array<int, int|string>>
+	 */
+	public function fileNotUploadedDataProvider() {
+		return [
+			['', 1],
+			['some name', 1],
+			['', 0],
+		];
+	}
```
SwikritiT, line 132: is this used?

individual-it: its a test to make sure the `if` check stays correct

SwikritiT: The data from this provider isn't used in the test function below.

individual-it: :facepalm: I've fixed it now

SwikritiT (approved):

PR #329, opened 2023-01-20
[OP#45849] when uploading the file check permissions correctly

individual-it:
When the there is no permission to create anything in the folder, return the correct error. But in the case of an overwrite the update permissions are enough.


SagarGi (commented):
Code LGTM :+1:

SwikritiT:
@individual-it this needs a rebase

individual-it:
@SwikritiT rebase done

SwikritiT (approved):

PR #338, opened 2023-01-24
put apache config into a volume

individual-it:
Having the apache config in a volume has the advantage that when the configuration is changed it will not be reset on `docker compose down/up`


SwikritiT (approved):

Kharonus (approved):
Works and looks fine!

PR #341, opened 2023-01-25
[OP#45466]Remove expired direct upload tokens using background jobs

SagarGi:
### Description
This PR implements the background jobs to remove the expired direct upload tokens from the databases.

### Related Workpackage
https://community.openproject.org/projects/817/work_packages/45466/activity

### DIFF
--- a/lib/Service/DatabaseService.php
+++ b/lib/Service/DatabaseService.php
@@ -31,7 +31,7 @@ class DatabaseService {
 	/**
 	 * @var IDBConnection
 	 */
-	private IDBConnection $db;
+	public IDBConnection $db;
 
 
 	/** @var string table name */
--- a/lib/Service/DatabaseService.php
+++ b/lib/Service/DatabaseService.php
@@ -24,7 +24,6 @@
 
 namespace OCA\OpenProject\Service;
 
-use DateTime;
 use OCP\DB\Exception;
 use OCP\IDBConnection;
 
@@ -109,10 +108,9 @@ class DatabaseService {
 	 */
 	public function deleteExpiredTokens(): void {
 		$query = $this->db->getQueryBuilder();
-		$date = new DateTime();
 		$query->delete($this->table)
 			->where(
-				$query->expr()->lt('expires_on', $query->createNamedParameter(($date)->getTimestamp()))
+				$query->expr()->lt('expires_on', $query->createNamedParameter(time()))
 			);
 		$query->execute();
 	}
--- a/lib/Service/DatabaseService.php
+++ b/lib/Service/DatabaseService.php
@@ -24,6 +24,7 @@
 
 namespace OCA\OpenProject\Service;
 
+use DateTime;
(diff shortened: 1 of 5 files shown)


lib/Service/DatabaseService.php
```diff
@@ -102,6 +103,27 @@ public function getTokenInfoFromDB(string $token): array {
...
+	 *
+	 * Get All token information from database
+	 *
+	 * @throws Exception
+	 */
+	public function deleteAllExpiredTokenFromDB(): void {
+		$query = $this->db->getQueryBuilder();
+		$query->select('token', 'expires_on')
+			->from($this->table);
+		$req = $query->executeQuery();
```
individual-it, line 116: just make a single request and let the DB deal with it, that will be much more performant that doing multiple requests.
MySQL should work like this: `DELETE FROM oc_direct_upload WHERE  expires_on < UNIX_TIMESTAMP();`

lib/Service/DirectUploadService.php
```diff
@@ -56,7 +56,6 @@ class DirectUploadService {
 
 	/** @var string time of token expiration */
 	private string $expiryTime = '+1 hour';
-
```
individual-it, line 59: not needed

lib/Service/DatabaseService.php
```diff
@@ -102,6 +103,27 @@ public function getTokenInfoFromDB(string $token): array {
 		];
 	}
 
+	/**
+	 *
+	 * Get All token information from database
```
individual-it, line 108: not true, be honest ;-)

lib/Service/DatabaseService.php
```diff
@@ -102,6 +103,27 @@ public function getTokenInfoFromDB(string $token): array {
...
+	 *
+	 * @throws Exception
+	 */
+	public function deleteAllExpiredTokenFromDB(): void {
```
individual-it, line 112:
```suggestion
	public function deleteExpiredTokens(): void {
```
as this is a DB class, its clear that its deleting from the DB

individual-it (changes requested):

lib/Service/DatabaseService.php
individual-it: for pgsql it might be easier to generate the timestamp in PHP. But still just have a single delete request

SagarGi: @individual-it do we want to use `UNIX_TIMESTAMP();` function which is for mysql. why not using the date instead? does it not work for pgsql?

individual-it:
The app is not enabled in CI when we run the unit tests. I would suggest to merge #323, rebase and then adjust then adjust the CI steps

individual-it (approved):

SagarGi:
> The app is not enabled in CI when we run the unit tests. I would suggest to merge #323, rebase and then adjust then adjust the CI steps

sure !!

SwikritiT (approved):

PR #344, opened 2023-01-26
use php 8.1 for release

individual-it:
NC master does not work with PHP7 anymore


wielinde (approved):
No idea, if it is good or not but this will unblock the merge, now :-)

PR #348, opened 2023-01-27
[OP#45941] Add more api tests for direct upload feature

SwikritiT:
related work package OP#45941 : https://community.openproject.org/projects/nextcloud-integration/work_packages/45941/activity
 
This PR
1. Adds API test
2. moves the `getTokenInfo` function call at the beginning as the token is deleted when this function is called. Previously, the token was usable if the request failed before it reached this function and we don't want that since the token is one-time only.

### DIFF
--- a/lib/Controller/DirectUploadController.php
+++ b/lib/Controller/DirectUploadController.php
@@ -38,6 +38,7 @@ use OCP\Files\InvalidPathException;
 use OCP\Files\NotEnoughSpaceException;
 use OCP\Files\NotFoundException;
 use OCA\OpenProject\Service\DirectUploadService;
+use OCA\OpenProject\Service\DatabaseService;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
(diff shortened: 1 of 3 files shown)


lib/Controller/DirectUploadController.php
```diff
@@ -145,6 +152,11 @@ public function prepareDirectUpload(int $folder_id): DataResponse {
 	 */
 	public function directUpload(string $token):DataResponse {
 		try {
+			if (strlen($token) !== 64 || !preg_match('/^[a-zA-Z0-9]*/', $token)) {
+				$this->databaseService->deleteToken($token);
+				throw new NotFoundException('invalid token');
+			}
+			$tokenInfo = $this->directUploadService->getTokenInfo($token);
```
SwikritiT, line 159: moving this here because the token is deleted when this function is called. Previously, the token was usable if the request failed before it reached this function and we don't want that since the token is one-time only.

SagarGi (commented):
LGTM :+1:

individual-it (approved):

PR #358, opened 2023-02-14
[OP#46177] create user and group openproject

SwikritiT:
## Related workpackages
OP#46177 -  https://community.openproject.org/projects/nextcloud-integration/work_packages/46177/activity
OP#46175 - https://community.openproject.org/projects/nextcloud-integration/work_packages/46175/activity

## Desciption
This PR implements following
1. Creates a user `openproject` that can neither be disabled or deleted
2. Creates a group `openproject` that cannot be deleted
3. Makes user `openproject` the admin of group `openproject`
4. Forbids the user `openproject` to logging through front end


individual-it:
The user still can be renamed and disabled, also the group can be disabled. We need to see if that also can be circumvented

individual-it (changes requested):
looks good to me, but let's merge on Monday

SagarGi (commented):
Looks Good to me. Checked it for integration script.

individual-it (approved):

PR #360, opened 2023-02-16
[OP#45928] hide the tab if the app is limited to specific groups

individual-it:
If the administrator limits the app to specific groups, don't show the OpenProject tab in the files details sidebar.

The limitation seems to work out of the box for all other places:
- user settings
- dashboard
- search
- navigation link
- APIs

fixes https://community.openproject.org/work_packages/45928

### DIFF
--- a/lib/Listener/LoadSidebarScript.php
+++ b/lib/Listener/LoadSidebarScript.php
@@ -29,6 +29,7 @@ namespace OCA\OpenProject\Listener;
 use OCA\Files\Event\LoadSidebar;
 use OCA\OpenProject\AppInfo\Application;
 use OCA\OpenProject\Service\OpenProjectAPIService;
+use OCP\App\IAppManager;
 use OCP\AppFramework\Services\IInitialState;
 use OCP\EventDispatcher\Event;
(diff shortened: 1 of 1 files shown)


SwikritiT:
@individual-it I think this needs a rebase with release/2.3

SwikritiT (changes requested):
Not working for me, I checked with nextcloud 24 and 25
1. Created a user 
2. added them to the group
3. limited that group
Screenshot from 2023-02-17 15-09-14

4. When I login with the user from that group I can still see the tab
Screenshot from 2023-02-17 15-09-51

SwikritiT (approved):
LGTM

SwikritiT:
Just one little question, if the admin creates a group and doesn't include themselves in the group and limit the usage to the group without them, the form isn't should for them and the set-up won't be possible. So do we add some check that the tab is always shown to the admin regardless they are in the group or not, or do we assume that the admin should be added to the group because that's the basic usage of the app?

individual-it:
I think we assume that the admin knows what to do. This PR anyway does not limit the settings in the admin panel, that happens automatically. We don't have control over that

PR #363, opened 2023-02-20
[OP#46179] protect project folders inside openproject group-folder from deletion

individual-it:


individual-it:
superseded by #364

PR #364, opened 2023-02-20
[OP#46179] create openproject group folder

individual-it:
Related workpackage OP#46179: https://community.openproject.org/projects/nextcloud-integration/work_packages/46179/activity

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     name: unit tests and linting
     strategy:
       matrix:
-        nextcloudVersion: [ stable22, stable23, stable24, stable25, stable26 ]
+        nextcloudVersion: [ stable22, stable23, stable24, stable25, stable26, master ]
         phpVersion: [ 7.4, 8.0, 8.1 ]
         exclude:
           - nextcloudVersion: stable22
@@ -22,8 +22,8 @@ jobs:
             phpVersion: 8.1
           - nextcloudVersion: stable26
             phpVersion: 7.4
-#          - nextcloudVersion: master
-#            phpVersion: 7.4
+          - nextcloudVersion: master
+            phpVersion: 7.4
     runs-on: ubuntu-latest
     steps:
       - name: Cancel previous runs
@@ -151,14 +151,14 @@ jobs:
         if: ${{ github.event_name == 'pull_request' && matrix.nextcloudVersion == 'master' && matrix.phpVersion == '8.1' }}
         uses: VeryGoodOpenSource/very_good_coverage@v1.2.0
         with:
-          min_coverage: '55'
+          min_coverage: '56.5'
           path: './server/apps/integration_openproject/coverage/php/lcov.info'
 
   api-tests:
     name: API tests
     strategy:
       matrix:
-        nextcloudVersion: [ stable22, stable23, stable24, stable25, stable26 ]
+        nextcloudVersion: [ stable22, stable23, stable24, stable25, stable26, master ]
         phpVersionMajor: [ 7, 8 ]
         phpVersionMinor: [ 4, 1 ]
         database: [pgsql, mysql]
@@ -180,14 +180,14 @@ jobs:
             phpVersionMinor: 1
           - phpVersionMajor: 8
             phpVersionMinor: 4
-#          - nextcloudVersion: master
-#            phpVersionMajor: 7
-#          - phpVersionMajor: 7
-#            phpVersionMinor: 0
-#          - phpVersionMajor: 7
-#            phpVersionMinor: 1
-#          - phpVersionMajor: 8
-#            phpVersionMinor: 4
+          - nextcloudVersion: master
+            phpVersionMajor: 7
+          - phpVersionMajor: 7
+            phpVersionMinor: 0
+          - phpVersionMajor: 7
+            phpVersionMinor: 1
+          - phpVersionMajor: 8
+            phpVersionMinor: 4
     runs-on: ubuntu-latest
     container: ubuntu:latest
     defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,7 +199,7 @@ jobs:
           SQL: ${{ matrix.database }}
           SERVER_BRANCH: ${{ matrix.nextcloudVersion }}
           NEXTCLOUD_AUTOINSTALL: "YES"
-          NEXTCLOUD_AUTOINSTALL_APPS: "viewer activity integration_openproject"
+          NEXTCLOUD_AUTOINSTALL_APPS: "viewer activity groupfolders integration_openproject"
           NEXTCLOUD_TRUSTED_DOMAINS: nextcloud
           VIRTUAL_HOST: "nextcloud"
           WITH_REDIS: "YES"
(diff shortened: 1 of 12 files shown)


SwikritiT:
` App "Group folders" cannot be installed because it is not compatible with this version of the server.`
https://github.com/nextcloud/integration_openproject/actions/runs/4352243691/jobs/7604789764#step:20:100 

Similar behaviour in my local machine too for the master of nc

SwikritiT:
https://github.com/nextcloud/groupfolders/pull/2287 the tests should pass after the merge of this PR

SwikritiT:
> One problem with the current implementation is that if I send the POST request to the `set-up` endpoint with `setup_group_folder = true` when any `openproject` entities are present the nextcloud `oauth credentials` aren't created Screenshot from 2023-03-13 14-39-23
> 
> This is because we create the oauth info
> 
> https://github.com/nextcloud/integration_openproject/blob/2d7790d86619fda4347ecf9889c56cfaacc22f96/lib/Controller/ConfigController.php#L519-L520
> 
> 
> after we call the `setIntegrationConfig`, which throws an exception if any open project entities are present

One workaround for this can be https://github.com/nextcloud/integration_openproject/pull/364/commits/8dd8407497ae069ec758898b3edcde853c700dd4

creating the NC oauth credentials when the `OpenprojectGroupfolderSetupConflictException`  exception occurs as we know this exception is due to some configuration related to the `OpenProject` entities.

lib/Service/OpenProjectAPIService.php
```diff
@@ -854,14 +889,141 @@ public function revokeUserOAuthToken(
...
+			$this->subAdminManager->isSubAdminOfGroup(
+				$this->userManager->get(Application::OPEN_PROJECT_ENTITIES_NAME),
+				$this->groupManager->get(Application::OPEN_PROJECT_ENTITIES_NAME)
+			)) {
+			return true;
+		}
+		return false;
+	}
+
+	public function isACLSetUp():bool {
```
individual-it, line 1011: this function checks if the OpenProject user has all permissions for the OpenProject group-folder. So maybe we can give it a clearer name e.g. `hasOpenProjectUserFullPermissions()`

I'm not sure if that will also return `false` if ACL is disabled all together. That could be checked e.g. like
```
	public function isACLEnabled(): bool {
		// @phpstan-ignore-next-line - make phpstan not complain if groupfolders app does not exist
		$groupfoldersFolderManager = $this->getGroupFolderManager();
		// @phpstan-ignore-next-line - make phpstan not complain if groupfolders app does not exist
		$folders = $groupfoldersFolderManager->getAllFolders();
		foreach ($folders as $folder) {
			if ($folder['mount_point'] === Application::OPEN_PROJECT_ENTITIES_NAME && $folder['acl'] === true) {
				return true;
			}
		}
		return false;
	}
```

There might be better ways to do that.

Also we need to check if the OpenProject user is allowed to manage the ACLs see `FolderManager::canManageACL()`

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -1217,14 +1267,184 @@ public function testRequestException(
...
+			->willReturn($folderManagerMock);
+		$service->method('isACLSetUp')->willReturn(true);
+		$this->assertTrue($service->isGroupFolderSetup());
+	}
+
+
+	/**
+	 * @return array<mixed>
+	 */
+	public function groupFolderNotSetUpDataProvider() {
```
individual-it, line 1360: what is the point of having the `$groupFolderPath` parameter if its the same in all test-cases?

SwikritiT: It's supposed to be different in one case. Thanks for pointing it out. I'll update it.

lib/Service/OpenProjectAPIService.php
```diff
@@ -854,14 +889,141 @@ public function revokeUserOAuthToken(
 	}
 
 	/**
-	 * @throws OpenprojectUserOrGroupAlreadyExistsException
+	 * @throws OpenprojectGroupfolderSetupConflictException
 	 */
 	public function isSystemReadyForGroupFolderSetUp(): bool {
 		if ($this->userManager->userExists(Application::OPEN_PROJECT_ENTITIES_NAME)) {
-			throw new OpenprojectUserOrGroupAlreadyExistsException('user "OpenProject" already exists');
+			throw new OpenprojectGroupfolderSetupConflictException('user "OpenProject" already exists');
```
SagarGi, line 896:
```suggestion
			throw new OpenprojectGroupfolderSetupConflictException("user " . Application::OPEN_PROJECT_ENTITIES_NAME .
					" already exists");
```

lib/Service/OpenProjectAPIService.php
```diff
@@ -854,14 +889,141 @@ public function revokeUserOAuthToken(
...
 		} elseif ($this->groupManager->groupExists(Application::OPEN_PROJECT_ENTITIES_NAME)) {
-			throw new OpenprojectUserOrGroupAlreadyExistsException('group "OpenProject" already exists');
+			throw new OpenprojectGroupfolderSetupConflictException('group "OpenProject" already exists');
```
SagarGi, line 898:
```suggestion
			throw new OpenprojectGroupfolderSetupConflictException('group' . Application::OPEN_PROJECT_ENTITIES_NAME .  ' already exists');
```

SwikritiT:
@individual-it @SagarGi this is ready for another round of review

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     name: unit tests and linting
     strategy:
       matrix:
-        nextcloudVersion: [ stable22, stable23, stable24, stable25, stable26, master ]
```
individual-it, line 14: I've merged https://github.com/nextcloud/groupfolders/pull/2287 so this can be reverted

.github/workflows/ci.yml
```diff
@@ -149,14 +151,14 @@ jobs:
         if: ${{ github.event_name == 'pull_request' && matrix.nextcloudVersion == 'master' && matrix.phpVersion == '8.1' }}
         uses: VeryGoodOpenSource/very_good_coverage@v1.2.0
         with:
-          min_coverage: '57'
+          min_coverage: '55'
```
individual-it, line 154: let's put it to 56 or even 56.5 because that is what we have currently
I think its OK if it decreased a bit, because a lot of the new code is tested by API tests

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -1217,14 +1269,230 @@ public function testRequestException(
...
+			->with(Application::OPEN_PROJECT_ENTITIES_NAME)
+			->willReturn($groupMock);
+		$groupManagerMock
+			->method('isInGroup')
+			->with(Application::OPEN_PROJECT_ENTITIES_NAME, Application::OPEN_PROJECT_ENTITIES_NAME)
+			->willReturn(true);
+
+		$folderMock = $this->getMockBuilder('\OCP\Files\Folder')->getMock();
+		$folderMock->
+		method('getFullPath')->
```
individual-it, line 1355: strange indent

individual-it (commented):
Looks very good to me, just some tiny comments

.github/workflows/ci.yml
SwikritiT: put it to `56.5`

tests/lib/Service/OpenProjectAPIServiceTest.php
SwikritiT: fixed it

SwikritiT:
> 

coverege increased. So maybe we can increase the min coverage too

PR #366, opened 2023-02-22
Translate error messages in direct upload

SwikritiT:
This PR translates the error messages.


individual-it (approved):

individual-it (changes requested):
this looks good to me, but let's merge on Monday

individual-it (approved):

PR #374, opened 2023-03-13
Update README

SwikritiT:


individual-it (commented):
we have the same thing saying at the top of the file
https://github.com/nextcloud/integration_openproject/pull/374/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L17
Should we keep both?

SwikritiT:
> we have the same thing saying at the top of the file https://github.com/nextcloud/integration_openproject/pull/374/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L17 Should we keep both?

I didn't realize we already had it. I'll close this PR

PR #381, opened 2023-03-31
String fixes

silopolis:
Small plurals in AdminSettings.vue component

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -578,12 +578,12 @@ export default {
 				break
 			case 'other_error':
 				showError(
-					t('integration_openproject', 'Failed to revoke some user(s) OpenProject OAuth access token(s)')
+					t('integration_openproject', 'Failed to revoke some users\' OpenProject OAuth access tokens')
 				)
 				break
 			case 'success':
 				showSuccess(
-					t('integration_openproject', 'Successfully revoked user(s) OpenProject OAuth access token(s)')
+					t('integration_openproject', 'Successfully revoked users\' OpenProject OAuth access tokens')
 				)
 				break
 			default:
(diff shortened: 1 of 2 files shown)


src/components/AdminSettings.vue
```diff
@@ -578,12 +578,12 @@ export default {
 				break
 			case 'other_error':
 				showError(
-					t('integration_openproject', 'Failed to revoke some user(s) OpenProject OAuth access token(s)')
+					t('integration_openproject', 'Failed to revoke some users\' OpenProject OAuth access tokens')
```
phil-davis, line 581: This is correct use of the apostrophe to indicate the possessive of a plural. There are "some users" (plural) and then we need to indicate the possessive of the plural "users" and so the apostrophe goes after the "s".

Ref: https://www.thepunctuationguide.com/apostrophe.html

phil-davis (approved):
LGTM

src/components/AdminSettings.vue
SwikritiT: what about 'tokens'?If there's only one token/

phil-davis: > what about 'tokens' ? If there's only one token?

IMO that is OK grammar. Example: "I created some files" - when "some" is used then the noun is plural "files" Usually when the word "some" is used we think "more than one", but I suppose there could be only one in a particular case.

If it is often the case that there is just one access token that failed to be revoked, then I suppose the code could try to detect that, and write a complete sentence in the singular.

SwikritiT:
Hi, @silopolis thank you for this PR. Some of the unit tests are failing because of this.
Can you please update with the same changes here as well
https://github.com/nextcloud/integration_openproject/blob/9f82042d7291b2120c8a7eeb78b5802afb14b70b/tests/jest/components/AdminSettings.spec.js#L953

https://github.com/nextcloud/integration_openproject/blob/9f82042d7291b2120c8a7eeb78b5802afb14b70b/tests/jest/components/AdminSettings.spec.js#L958

silopolis:
Of course, will do later today 👍

silopolis:
A bit late, but done :)

individual-it (approved):

SwikritiT:
@individual-it maybe we can merge this? The tests are passing
https://github.com/nextcloud/integration_openproject/pull/381,
CI is red because of the coverage cannot be posted from a fork.

PR #385, opened 2023-04-03
Make files-info endpoint return path and permissions of the file

SwikritiT:
Related workpackages:
- [OP#47266] - https://community.openproject.org/projects/nextcloud-integration/work_packages/47266/activity
- [OP#47260] - https://community.openproject.org/projects/nextcloud-integration/work_packages/47260

### DIFF
--- a/lib/Controller/FilesController.php
+++ b/lib/Controller/FilesController.php
@@ -216,7 +216,7 @@ class FilesController extends OCSController {
 				);
 				$internalPath = $file->getName();
 			}
-
+			$path = $file->getInternalPath();
 			$modifier = $this->getLastModifier($owner->getUID(), $file->getId());
 			if ($modifier instanceof IUser) {
 				$modifierId = $modifier->getUID();
@@ -246,7 +246,7 @@ class FilesController extends OCSController {
 				'modifier_name' => $modifierName,
 				'modifier_id' => $modifierId,
 				'dav_permissions' => DavUtil::getDavPermissions($file),
-				'path' => $internalPath
+				'path' => $path,
 			];
 		}
 
--- a/lib/Controller/FilesController.php
+++ b/lib/Controller/FilesController.php
@@ -35,6 +35,7 @@ use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\RichObjectStrings\IValidator;
 use Throwable;
+use OCP\Files\DavUtil;
 
 class FilesController extends OCSController {
 
@@ -244,7 +245,7 @@ class FilesController extends OCSController {
 				'trashed' => $trashed,
 				'modifier_name' => $modifierName,
 				'modifier_id' => $modifierId,
-				'dav_permissions' => DavUtil::_getDavPermissions_($file),
+				'dav_permissions' => DavUtil::getDavPermissions($file),
 				'path' => $internalPath
 			];
 		}
--- a/lib/Controller/FilesController.php
+++ b/lib/Controller/FilesController.php
@@ -35,7 +35,6 @@ use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\RichObjectStrings\IValidator;
 use Throwable;
-use OCP\Files\DavUtil;
 
 class FilesController extends OCSController {
 
@@ -233,8 +232,7 @@ class FilesController extends OCSController {
 			// full path is in format `<user-name>/files/a/b/`
 			// since we don't want to send it with username, only get the `files/a/b` and send it
 			$path = explode('/', $fullpath, 3);
-			// @phpstan-ignore-next-line
-			$davPermission = DavUtil::getDavPermissions($file);
+			$davPermission = $this->getDavPermissions($file);
 			return [
 				'status' => 'OK',
 				'statuscode' => 200,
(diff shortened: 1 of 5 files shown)


lib/Controller/FilesController.php
```diff
@@ -243,7 +244,9 @@ private function compileFileInfo($fileId) {
 				'owner_id' => $owner->getUID(),
 				'trashed' => $trashed,
 				'modifier_name' => $modifierName,
-				'modifier_id' => $modifierId
+				'modifier_id' => $modifierId,
+				'dav_permissions' => DavUtil::getDavPermissions($file),
+				'path' => $internalPath
```
Kharonus, line 249: taking code from 204, `$internalPath` is most of the time the path of the owner, not the requesting user. But the later one is the one, we need.

Kharonus (changes requested):

lib/Controller/FilesController.php
SwikritiT: true. I'll change it

SwikritiT: @Kharonus do you need it like `<user>/files/a/b/..` or `files/a/a/..` or does it not matter?

Kharonus: @SwikritiT Thanks for changing it. 

It should not matter, if the user is prefixed to the location path or not. We shift around values without the user, so we would have lesser postprocessing if the location is `files/a/b`. What is your opinion on that?

SwikritiT: Yes, makes sense. I've returned the one without the user.

Kharonus (approved):

lib/Controller/FilesController.php
```diff
@@ -229,7 +229,12 @@ private function compileFileInfo($fileId) {
 			} else {
 				$mimeType = $file->getMimeType();
 			}
-
+			$fullpath = $file->getpath();
```
SwikritiT, line 232: We cannot use `getInternalPath()` here like in https://github.com/nextcloud/integration_openproject/pull/385/commits/7fa31074f63c4844c7821f285f073dd3665d8d1c because `getInternalPath()` gets the path relative to the storage, and in case of a shared resource, the request from the share receiver will get an empty internal path as they don't own the file and the storage is that of the owner.

SwikritiT:
Seems like this file https://github.com/nextcloud/server/blob/stable25/lib/public/Files/DavUtil.php is only available from nc-25 😢 , need to find some alternative for support in NC 24 and 23

lib/Controller/FilesController.php
```diff
@@ -316,4 +321,38 @@ class_exists('\OCA\Activity\UserSettings')
 		}
 		return null;
 	}
+
+	// copy of https://github.com/nextcloud/server/blob/cd44ee2053f105f60c668906ed8460605cb1da81/lib/public/Files/DavUtil.php#L58
```
individual-it, line 325: can we do `$file->getDavPermissions()` for the older versions? as in https://github.com/nextcloud/server/blame/e34f2c4799b271ac1b1c412af4bf2fa1f6279359/apps/dav/lib/BulkUpload/BulkUploadPlugin.php#L100

tests/acceptance/features/api/getFileinfoByFileIDAPI.feature
```diff
@@ -37,7 +39,9 @@ Feature: retrieve file information of a single file, using the file ID
           "owner_name": {"type": "string", "pattern": "^Alice$"},
           "modifier_id": {"type": "null"},
           "modifier_name": {"type": "null"},
-          "trashed": {"type": "boolean", "enum": [false]}
+          "trashed": {"type": "boolean", "enum": [false]},
+          "dav_permissions": {"type": "string", "pattern":"^RGDNVW$"},
+          "path": {"type": "string", "pattern":"^files/file.txt$"}
```
individual-it, line 44: I think you are stripping away `files` in https://github.com/nextcloud/integration_openproject/pull/385/files#diff-776f8feca78c8322be8a9c5a0a5c6dcb93dfdd2776045ee36faa1a4c172d645fR234
```suggestion
          "path": {"type": "string", "pattern":"^file.txt$"}
```

tests/acceptance/features/api/getFileinfoByFileIDAPI.feature
```diff
@@ -167,7 +181,9 @@ Feature: retrieve file information of a single file, using the file ID
           "owner_name": {"type": "string", "pattern": "^Alice$"},
           "modifier_id": {"type": "null"},
           "modifier_name": {"type": "null"},
-          "trashed": {"type": "boolean", "enum": [true]}
+          "trashed": {"type": "boolean", "enum": [true]},
+          "dav_permissions": {"type": "string", "pattern":"^RGDNVW$"},
+          "path": {"type": "string", "pattern":"^files_trashbin\/files\/subfolder.d\\d{10}\/file.txt"}
```
individual-it, line 186: what to do with that path? Strip away `files_trashbin/files` or leave both?

individual-it (changes requested):
I think we need more tests where we share with different permissions

tests/acceptance/features/api/getFileinfoByFileIDAPI.feature
SwikritiT: no the path is `<username>/files/a/b/c>` I'm stripping the `username` so it's `files/a/b/c`

tests/acceptance/features/api/getFileinfoByFileIDAPI.feature
SwikritiT: the internal path starts with `files_trashbin/` so I don't think we need to strip that away.
We can if @Kharonus doesn't need that info.

lib/Controller/FilesController.php
SwikritiT: I'll try it. Thanks!

SwikritiT: I tried but it didn't work. that function needs the node to be an instance of Sabre Node but it our case it's not

individual-it (approved):

PR #390, opened 2023-04-13
[OP#47630] Check if free space is a positive value

SwikritiT:
Currently we're checking quota like this

```php
$freeSpace = $folderNode->getFreeSpace();

// this is also true if we try to overwrite
// to overwrite a file we need enough free quota for the new data
// otherwise `putContent()` fails
if ($directUploadFile['size'] > $freeSpace) {
	throw new NotEnoughSpaceException('insufficient quota');
}
```
But since there's a possibility that the function getFreeSpace might return a negative

https://github.com/nextcloud/server/blob/b1abc57c07a61b5a0557395d86c3e31486a96c23/lib/public/Files/FileInfo.php#L52

https://github.com/nextcloud/server/blob/b1abc57c07a61b5a0557395d86c3e31486a96c23/lib/public/Files/FileInfo.php#L57

https://github.com/nextcloud/server/blob/b1abc57c07a61b5a0557395d86c3e31486a96c23/lib/public/Files/FileInfo.php#L62

So treat a negative return as an unlimited space

Related work package: https://community.openproject.org/projects/nextcloud-integration/work_packages/47630/activity?query_id=3504

### DIFF
--- a/lib/Controller/DirectUploadController.php
+++ b/lib/Controller/DirectUploadController.php
@@ -30,6 +30,7 @@ use OC\User\NoUserException;
 use InvalidArgumentException;
 use OC\ForbiddenException;
 use OCA\OpenProject\Exception\OpenprojectFileNotUploadedException;
+use OCA\OpenProject\Exception\OpenprojectUnauthorizedUserException;
 use \OCP\AppFramework\ApiController;
 use OCP\Files\File;
 use OCP\Files\InvalidCharacterInPathException;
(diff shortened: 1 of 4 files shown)


individual-it (changes requested):
looks good, can we add unit tests for that case?

SwikritiT:
> looks good, can we add unit tests for that case?

yes I can add it

lib/Controller/DirectUploadController.php
```diff
@@ -279,6 +283,14 @@ public function directUpload(string $token):DataResponse {
 			return new DataResponse([
 				'error' => $this->l->t($e->getMessage())
 			], Http::STATUS_UNSUPPORTED_MEDIA_TYPE);
+		} catch (GenericFileException $e) {
+			return new DataResponse([
+				'error' => $this->l->t($e->getMessage())
+			], Http::STATUS_BAD_REQUEST);
```
individual-it, line 289: In what case would that happen?
The response code does indicate that the client has done some mistake.

SwikritiT: This will be thrown when if the put file contents operation fails https://github.com/nextcloud/server/blob/9db33055b208518601ff7b389188c032f8792e39/lib/private/Files/Node/File.php#L71
What would be an appropriate error code for this?

SwikritiT: Removed it so that general exception will catch this as an internal server error.

PR #380, opened 2023-03-30
Update some npm pkgs and make adjustments

julien-nc:
@nextcloud/axios @nextcloud/eslint-config @nextcloud/router @nextcloud/stylelint-config @nextcloud/vue @nextcloud/webpack-vue-config vue vue-material-design-icons @nextcloud/dialogs and vue-template-compiler were updated.
Some small adjustments were needed.

@vue/cli-plugin-unit-jest was not updated, just in case.
Anyway the jest snapshots probably need to be regenerated.

PR #399, opened 2023-04-27
[OP#48106] Add a reference provider and a reference widget for work packages

julien-nc:
This is just the structure, the provider and the widget do nothing at the moment.
This requires https://github.com/nextcloud/integration_openproject/pull/380 as the widget registration function appeared recently in `@nextcloud/vue` and the latest `@nextcloud/webpack-vue-config` is required.

The registration of the reference provider and of the widget only happen on NC >= 26.

The reference system is documented there: https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/reference.html

The reference provider has 2 goals:
* Add a smart picker provider so users can get work package links
* Resolve and render work package link previews

### The smart picker

There are 2 ways to implement a smart picker provider. This PR implements the first one.

1. Just use a unified search provider and let the picker show a generic rendering for the provider (like for other apps like integration_github, integration_gitlab, collectives, deck, tables...). In this case there is nothing to implement on the frontend side. Users can perform a basic search and the picker result is the "resourceUrl" of the selected search result entry.
2. It is possible to register a custom picker component to have a custom interface. For example, this is done in integration_openstreetmap (it shows a map to get a location link), in integration_giphy (it shows a GIF picker with a grid view and infinite scrolling) etc...

We can start by implementing the first one and see if it makes sense to have a custom picker later. It would allow us to add search options for example.

### The link previews (reference widgets)

If a reference provider recognizes a link that it can handle, it resolves it and provides information for the frontend.
If only the generic title, description and image are set, the default rendering will be used.
The rendering can be customized if a rich object is set and a widget component is registered for this rich object type.
This is done in this PR.

Related workpackage: https://community.openproject.org/projects/nextcloud-integration/work_packages/48106/activity

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -15,5 +15,6 @@ $config
 	->notPath('src')
 	->notPath('vendor')
 	->notPath('server')
+	->notPath('lib/Reference/WorkPackageReferenceProvider.php')
 	->in(__DIR__);
 return $config;
--- a/lib/Reference/WorkPackageReferenceProvider.php
+++ b/lib/Reference/WorkPackageReferenceProvider.php
@@ -1,0 +1,188 @@
+<?php
+
+/**
+ * @copyright Copyright (c) 2023 Julien Veyssier <julien-nc@posteo.net>
+ *
+ * @author Julien Veyssier <julien-nc@posteo.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCA\OpenProject\Reference;
+
+use OCA\OpenProject\Service\OpenProjectAPIService;
+use OCP\Collaboration\Reference\ADiscoverableReferenceProvider;
+use OCP\Collaboration\Reference\ISearchableReferenceProvider;
+use OCP\Collaboration\Reference\Reference;
+use OC\Collaboration\Reference\ReferenceManager;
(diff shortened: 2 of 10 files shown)


SwikritiT:
https://github.com/nextcloud/integration_openproject/pull/380 is merged now, I'll rebase and look around.

lib/Reference/WorkPackageReferenceProvider.php
```diff
@@ -0,0 +1,188 @@
...
+
+		return null;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function matchReference(string $referenceText): bool {
+		if ($this->userId !== null) {
+			$linkPreviewEnabled = $this->config->getUserValue($this->userId, Application::APP_ID, 'default_enable_unified_search', '0') === '1';
```
SwikritiT, line 123: reminder to self to change these and  add a migration step to add this key
```suggestion
			$linkPreviewEnabled = $this->config->getUserValue($this->userId, Application::APP_ID, ''link_preview_enabled'', '0') === '1';
```

lib/Reference/WorkPackageReferenceProvider.php
```diff
@@ -0,0 +1,188 @@
...
+			if (!$linkPreviewEnabled) {
+				return false;
+			}
+		}
+		$adminLinkPreviewEnabled = $this->config->getAppValue(Application::APP_ID, 'default_enable_unified_search', '0') === '1';
```
SwikritiT, line 128:
```suggestion
		$adminLinkPreviewEnabled = $this->config->getAppValue(Application::APP_ID, ''link_preview_enabled'', '0') === '1';
```

phpstan.neon
```diff
@@ -7,3 +7,5 @@ parameters:
         - bootstrap.php
     checkGenericClassInNonGenericObjectType: false
     reportUnmatchedIgnoredErrors: false
+    excludePaths:
+    	-  './lib/Reference/WorkPackageReferenceProvider.php'
```
SwikritiT, line 11: Phpstan will throw `Reflection error: OCP\Collaboration\Reference\ISearchableReferenceProvider not found.` in nextcloud server version lower that 25, as this is not available in lower versions, so exclude the path

individual-it (approved):
Looks good to me, let's merge it when CI passes

.php-cs-fixer.dist.php
```diff
@@ -15,5 +15,6 @@
 	->notPath('src')
 	->notPath('vendor')
 	->notPath('server')
+	->notPath('lib/Reference/WorkPackageReferenceProvider.php')
```
SwikritiT, line 18: similar to this https://github.com/nextcloud/integration_openproject/pull/399/files#r1230459632 but with `cs:fixer`

PR #408, opened 2023-05-11
Add environment variable for retrying auto installation

SwikritiT:
This PR https://github.com/juliushaertl/nextcloud-docker-dev/pull/189 made the auto retry of app installation default to 0 so we add the environment variable to `120`.

### DIFF
--- a/src/views/ProjectsTab.vue
+++ b/src/views/ProjectsTab.vue
@@ -321,8 +321,8 @@ export default {
 		}
 		&--unlinkactionbutton {
 			position: absolute;
-			right:0;
-			top:0;
+			right: 0;
+			top: 0;
 			margin: 4px;
 			visibility: hidden;
 		}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,6 +198,7 @@ jobs:
           NEXTCLOUD_TRUSTED_DOMAINS: nextcloud
           VIRTUAL_HOST: "nextcloud"
           WITH_REDIS: "YES"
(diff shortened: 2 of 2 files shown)


src/views/ProjectsTab.vue
```diff
@@ -321,8 +321,8 @@ export default {
 		}
 		&--unlinkactionbutton {
 			position: absolute;
-			right:0;
-			top:0;
+			right: 0;
+			top: 0;
```
individual-it, line 325: this change does not look like it belongs into this PR

individual-it (changes requested):

src/views/ProjectsTab.vue
SwikritiT: I don't know why the stylelint of the CI of previous PR https://github.com/nextcloud/integration_openproject/pull/407 didn't complain about this issue but in this PR https://github.com/nextcloud/integration_openproject/actions/runs/4945800995/jobs/8843015517 it complained so the CI might be red.

SwikritiT: I've created a seperated PR for this https://github.com/nextcloud/integration_openproject/pull/409. I'll rebase once it's merged

individual-it (approved):

SwikritiT:
I'm merging this because this https://github.com/nextcloud/integration_openproject/pull/409 won't pass the CI as it needs the variable added by this PR because activity app is not installed without it https://github.com/nextcloud/integration_openproject/actions/runs/4955244949/jobs/8867408836?pr=409#step:18:94

PR #411, opened 2023-05-18
[OP#48255]Fix global search to display all the work packages available to the user

SwikritiT:
Related work package [OP#48255] : https://community.openproject.org/projects/nextcloud-integration/work_packages/48255
Currently, the global search of the nextcloud search only searches for the work packages that are linkable by the user. This PR fixes it to return all the work packages that are available to the user

needs https://github.com/nextcloud/integration_openproject/pull/410 to be merged first

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -213,7 +213,7 @@ class OpenProjectAPIService {
 	 * @param string $userId
 	 * @param string|null $query
 	 * @param int|null $fileId
-	 * @param bool $isGlobalSearch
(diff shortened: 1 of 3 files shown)


lib/Service/OpenProjectAPIService.php
```diff
@@ -222,8 +222,7 @@ public function searchWorkPackage(
 		string $userId,
 		string $query = null,
 		int $fileId = null,
-		int $offset = 0,
-		int $limit = 5
```
SwikritiT, line 226: these weren't used so removed them

individual-it (approved):

PR #414, opened 2023-05-25
rely on foldername not fileID for delete protection of project folders

individual-it:
Relying on the fileID might be problematic because the admin might have deleted the OpenProject groupfolder and recreated it, in that case the fileID would have changed, but we would not know about it


SwikritiT (approved):
LGTM

SagarGi (commented):
LGTM :+1:

SwikritiT (approved):

PR #415, opened 2023-05-25
improve check if the groupfolder is setup correctly

individual-it:
Using the storage and the userfolder to check the groupfolder configuration seems to be bridle, because the user called OpenProject might not be completely finalized (first login not done) when we call this check. We didn't manage to finalize the login programmatically in a way, so that the groupfolders app would create the mounts etc, sometimes we are getting lock exceptions.
But luckily we don't need to rely on the userfolder, all the checks that we want, we can do through the `groupFolderManager`

This PR changes the code so that we have one function that checks:
 - if the group names OpenProject is managing a folder called OpenProject
 - if the permissions are set correctly
 - if the ACL is enabled and can be managed by the user OpenProject
It also renames the function to make clear what is about the configuration of the groupfolder app and what is about the complete project folder setup


SwikritiT (approved):
LGTM

PR #416, opened 2023-05-26
Replace `master` branch in ci with `stable27`

SwikritiT:
The master of nextcloud is now 28 so we'll have a error in CI saying that this app is not compatible for master, so change the master to `stable27`


individual-it (changes requested):
We need to change all occurrences of "master" in the release process

individual-it (approved):

PR #422, opened 2023-06-07
[OP#48398] import self-signed certificates from OpenProject

individual-it:
If the OpenProject instance runs with a self-signed certificated, the admin should have an option to accept it


individual-it:
closing, because we have decided not to do that

PR #425, opened 2023-06-09
[OP#48445]Update integration script according to project folder setup

SagarGi:
### Description
This PR adjust the bash script for whole set up integration since now we have an option to set the integration.
1. with project folder setup
2. without project folder setup

This PR adds and extra environment `SETUP_PROJECT_FOLDER` where admin can set the with or without the project folder setup. 

### Related work package:
https://community.openproject.org/projects/nextcloud-integration/work_packages/48445/activity?query_id=3787

### DIFF
--- a/integration_setup.sh
+++ b/integration_setup.sh
@@ -13,6 +13,8 @@
 # OPENPROJECT_STORAGE_NAME=<openproject_filestorage_name> This variable is the name of the storage which keeps the oauth information in OpenProject required for integration.
 # SETUP_PROJECT_FOLDER= <true|false> If true set the integration is done with project folder setup and vice-versa
 
+
+# helper functions
 log_error() {
 	echo -e "\e[31m$1\e[0m"
 }
@@ -23,6 +25,18 @@ log_info() {
 
 log_success() {
 	echo -e "\e[32m$1\e[0m"
+}
+
+getErrorLogSettingOPOauthInfoToNextcloud() {
+	echo  "$1" | jq
+	log_error "Setup of the integration failed :( !!"
+	# when there is error when saving OP oauth information in the nextcloud
+	# we can delete the storage created in openproject, otherwise it would need to be deleted manually when the script is ran the next time
+	curl -s -X DELETE -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
+										  ${OPENPROJECT_BASEURL_FOR_STORAGE}/${storage_id} \
+										  -H 'accept: application/hal+json' \
+										  -H 'Content-Type: application/json' \
+										  -H 'X-Requested-With: XMLHttpRequest'
 }
 
 # making sure that jq is installed
@@ -153,23 +167,11 @@ nextcloud_information_response=$(curl -s -X${setup_method} -u${NC_ADMIN_USERNAME
 nextcloud_client_id=$(echo $nextcloud_information_response | jq -e ".nextcloud_client_id")
 nextcloud_client_secret=$(echo $nextcloud_information_response | jq -e ".nextcloud_client_secret")
 
-setOPInfoInNextcloudErrorLog() {
-	echo "${nextcloud_information_response}" | jq
-	log_error "Setup of the integration failed :( !!"
-	# when there is error when saving OP oauth information in the nextcloud
-	# we can delete the storage created in openproject, otherwise it would need to be deleted manually when the script is ran the next time
-	curl -s -X DELETE -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
-										  ${OPENPROJECT_BASEURL_FOR_STORAGE}/${storage_id} \
-										  -H 'accept: application/hal+json' \
-										  -H 'Content-Type: application/json' \
-										  -H 'X-Requested-With: XMLHttpRequest'
-}
-
 if [[ ${SETUP_PROJECT_FOLDER} == true ]]
 then
 	openproject_user_app_password=$(echo $nextcloud_information_response | jq -e ".openproject_user_app_password")
 	if [ ${openproject_user_app_password} == null ]; then
-		setOPInfoInNextcloudErrorLog
+		getErrorLogSettingOPOauthInfoToNextcloud "$nextcloud_information_response"
 		log_info "Above response is missing one or more of nextcloud_client_id, nextcloud_client_secret, or openproject_user_app_password"
 		log_info "If the error response is related to project folder setup name 'OpenProject' (group, user, folder) then follow below link to resolve the error"
 		log_info "Visit this link to resolve the error manually https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/"
@@ -177,7 +179,7 @@ then
 	fi
 else
 	if [ ${nextcloud_client_id} == null ] || [ ${nextcloud_client_secret} == null ]; then
-    	setOPInfoInNextcloudErrorLog
+		getErrorLogSettingOPOauthInfoToNextcloud "$nextcloud_information_response"
         log_info "Above response is missing nextcloud_client_id or nextcloud_client_secret"
         exit 1
     fi
--- a/integration_setup.sh
+++ b/integration_setup.sh
@@ -32,11 +32,17 @@ getErrorLogSettingOPOauthInfoToNextcloud() {
 	log_error "Setup of the integration failed :( !!"
 	# when there is error when saving OP oauth information in the nextcloud
 	# we can delete the storage created in openproject, otherwise it would need to be deleted manually when the script is ran the next time
-	curl -s -X DELETE -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
+	delete_storage_response=$(curl -s -o /dev/null -w "%{http_code}" -X DELETE -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
 										  ${OPENPROJECT_BASEURL_FOR_STORAGE}/${storage_id} \
 										  -H 'accept: application/hal+json' \
 										  -H 'Content-Type: application/json' \
-										  -H 'X-Requested-With: XMLHttpRequest'
+										  -H 'X-Requested-With: XMLHttpRequest')
+	if [[ ${delete_storage_response} == 204 ]]
+    then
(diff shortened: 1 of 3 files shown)


SagarGi:
This script for now when the `project folder is already setup` then it throws error when the user tries to set up it again. May be we can make it more user or admin friendly by adding checks for project folder setup. Need some Opinion regarding it.
CC @SwikritiT @individual-it

integration_setup.sh
```diff
@@ -53,6 +58,26 @@ then
 	exit 1
 fi
 
+# we can set whether we want the integration with project folder or without it using environment variable 'SET_PROJECT_FOLDER_SETUP'
+if [[ ${SET_PROJECT_FOLDER_SETUP} == true ]]
+then
+	# make an api request to get the status of the project folder setup
+	project_folder_setup_response=$(curl -s -X GET -u${NC_ADMIN_USERNAME}:${NC_ADMIN_PASSWORD} \
+								${NC_INTEGRATION_BASE_URL}/project-folder-status \
```
SwikritiT, line 66: should we also add `-k` in all of the curl requests? I have open project setup in HTTPS and I cannot use this without `-k` option
```suggestion
	project_folder_setup_response=$(curl -k -s -X GET -u${NC_ADMIN_USERNAME}:${NC_ADMIN_PASSWORD} \
								${NC_INTEGRATION_BASE_URL}/project-folder-status \
```

SwikritiT (commented):
@SagarGi the available env variable section should also be updated at the top to include `SET_PROJECT_FOLDER_SETUP` option

lib/Controller/ConfigController.php
```diff
@@ -595,20 +595,21 @@ public function updateIntegration(array $values): DataResponse {
...
-				}
-				if ($status['oPUserAppPassword'] !== null) {
-					$result['openproject_user_app_password'] = $status['oPUserAppPassword'];
-				}
-				return new DataResponse($result);
+				$result = array_merge($this->oauthService->getClientInfo($id), $result);
+			} else {
+				// we will recreate new oauth when admin has reset it
+				$result = array_merge($this->recreateOauthClientInformation(), $result);
 			}
```
SagarGi, line 611: I have updated this part. In case project folder is already setup then the set up is done with `PATCH`. It means that when the integration is reset and no  there is no oauthClient, the `PATCH` request creates the new oathClient informaton. Needs some opinion regarding this since PATCH request is related to update.

integration_setup.sh
```diff
@@ -39,7 +40,12 @@ NEXTCLOUD_HOST_MAINTENANCE_STATE=$(echo $NEXTCLOUD_HOST_STATE | jq -r ".maintena
 OPENPROJECT_HOST_STATE=$(curl -s -X GET -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
                           ${OPENPROJECT_HOST}/api/v3/configuration | jq -r "._type")
 OPENPROJECT_BASEURL_FOR_STORAGE=${OPENPROJECT_HOST}/api/v3/storages
-INTEGRATION_URL_FOR_SETUP=${NEXTCLOUD_HOST}/index.php/apps/integration_openproject/setup
+NC_INTEGRATION_BASE_URL=${NEXTCLOUD_HOST}/index.php/apps/integration_openproject
+# these two data are set as false when the integration is done without project folder setup
```
individual-it, line 44:
```suggestion
# these two data are set to "false" when the integration is done without project folder setup
```

integration_setup.sh
```diff
@@ -53,9 +59,29 @@ then
...
+    else
+    	setup_project_folder=true
+        setup_app_password=true
+	fi
+fi
+
 log_info "Creating file storage in OpenProject ..."
 # api call to get openproject_client_id and openproject_client_secret
-create_storage_response=$(curl -s -X POST -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
+create_storage_response=$(curl -k -s -X POST -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
```
individual-it, line 84: `-k` should not be used here because it will make the requests insecure

integration_setup.sh
```diff
@@ -124,18 +151,45 @@ nextcloud_information_response=$(curl -s -XPOST -u${NC_ADMIN_USERNAME}:${NC_ADMI
...
 	log_error "Setup of the integration failed :( !!"
+	if [ ${SET_PROJECT_FOLDER_SETUP} == true ]; then
+		log_info "Above response does not contain nextcloud_client_id or nextcloud_client_secret or openproject_user_app_password"
+		log_info "If the error is related to project folder setup name 'OpenProject' (group, user, folder) then follow below link to resolve the error"
+		echo "Visit this link to resolve the error manually https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/"
+	else
+		log_info "Above response does not contain nextcloud_client_id or nextcloud_client_secret"
+	fi
+	# when there is error when saving OP oauth information in the nextcloud
+	# we can delete the storage created in openproject. since we have to delete it manually when we run the script for the next time
```
individual-it, line 165:
```suggestion
	# we can delete the storage created in openproject, otherwise it would need to be deleted manually when the script is ran the next time
```

integration_setup.sh
```diff
@@ -53,9 +59,29 @@ then
 	exit 1
 fi
 
+# we can set whether we want the integration with project folder or without it using environment variable 'SET_PROJECT_FOLDER_SETUP'
+if [[ ${SET_PROJECT_FOLDER_SETUP} == true ]]
```
individual-it, line 63:
```suggestion
if [[ ${SETUP_PROJECT_FOLDER} == true ]]
```

integration_setup.sh
```diff
@@ -39,7 +40,12 @@ NEXTCLOUD_HOST_MAINTENANCE_STATE=$(echo $NEXTCLOUD_HOST_STATE | jq -r ".maintena
 OPENPROJECT_HOST_STATE=$(curl -s -X GET -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
                           ${OPENPROJECT_HOST}/api/v3/configuration | jq -r "._type")
 OPENPROJECT_BASEURL_FOR_STORAGE=${OPENPROJECT_HOST}/api/v3/storages
-INTEGRATION_URL_FOR_SETUP=${NEXTCLOUD_HOST}/index.php/apps/integration_openproject/setup
+NC_INTEGRATION_BASE_URL=${NEXTCLOUD_HOST}/index.php/apps/integration_openproject
+# these two data are set as false when the integration is done without project folder setup
+setup_project_folder=false
+setup_app_password=false
+setup_method=POST
```
individual-it, line 47: this is only used once, so why not calculate the value where ever its used

individual-it (changes requested):

integration_setup.sh
SagarGi: The value changes from post to patch when project folder setup is already set up and also the endpoint is the same.

integration_setup.sh
```diff
@@ -124,13 +151,40 @@ nextcloud_information_response=$(curl -s -XPOST -u${NC_ADMIN_USERNAME}:${NC_ADMI
 nextcloud_client_id=$(echo $nextcloud_information_response | jq -e ".nextcloud_client_id")
 nextcloud_client_secret=$(echo $nextcloud_information_response | jq -e ".nextcloud_client_secret")
 
-if [ ${nextcloud_client_id} == null ] || [ ${nextcloud_client_secret} == null ]; then
+setOPInfoInNextcloudErrorLog() {
 	echo "${nextcloud_information_response}" | jq
-	log_error "Response does not contain nextcloud_client_id (nextcloud_client_id) or nextcloud_client_secret(nextcloud_client_secret)"
 	log_error "Setup of the integration failed :( !!"
+	if [ ${SETUP_PROJECT_FOLDER} == true ]; then
+		log_info "Above response does not contain nextcloud_client_id or nextcloud_client_secret or openproject_user_app_password"
```
individual-it, line 158:
```suggestion
		log_info "Above response is missing one or more of nextcloud_client_id, nextcloud_client_secret, or openproject_user_app_password"
```

integration_setup.sh
```diff
@@ -124,13 +151,40 @@ nextcloud_information_response=$(curl -s -XPOST -u${NC_ADMIN_USERNAME}:${NC_ADMI
...
+		log_info "If the error is related to project folder setup name 'OpenProject' (group, user, folder) then follow below link to resolve the error"
+		echo "Visit this link to resolve the error manually https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/"
+	else
+		log_info "Above response does not contain nextcloud_client_id or nextcloud_client_secret"
```
individual-it, line 162:
```suggestion
		log_info "Above response is missing nextcloud_client_id or nextcloud_client_secret"
```

integration_setup.sh
```diff
@@ -107,15 +133,16 @@ log_info "success!"
 
 log_info "Setting up OpenProject integration in Nextcloud..."
 # api call to set the  openproject_client_id and openproject_client_secret to Nextcloud and also get nextcloud_client_id and nextcloud_client_secret
-nextcloud_information_response=$(curl -s -XPOST -u${NC_ADMIN_USERNAME}:${NC_ADMIN_PASSWORD} ${INTEGRATION_URL_FOR_SETUP} \
+nextcloud_information_response=$(curl -s -X${setup_method} -u${NC_ADMIN_USERNAME}:${NC_ADMIN_PASSWORD} "${NC_INTEGRATION_BASE_URL}/setup" \
```
individual-it, line 136:
```suggestion
if [ "$isProjectFolderAlreadySetup" == "true" ]; then
        setup_method=PATCH
else
        setup_method=POST
fi
nextcloud_information_response=$(curl -s -X${setup_method} -u${NC_ADMIN_USERNAME}:${NC_ADMIN_PASSWORD} "${NC_INTEGRATION_BASE_URL}/setup" \
```

lib/Controller/ConfigController.php
individual-it: I think that is OK

individual-it (changes requested):

integration_setup.sh
individual-it: maybe do it like that: https://github.com/nextcloud/integration_openproject/pull/425/files#r1246144808

SwikritiT (approved):
LGTM

PR #426, opened 2023-06-19
[OP#48397]Show navigation link to admin if integration setup is not done

SwikritiT:
Related workpackage[OP#48397] : https://community.openproject.org/projects/nextcloud-integration/work_packages/48397

In the personal section of a user, if the user is an admin, show them navigation to the admin settings if the setup seems to be not complete. For the normal user the message is not changed.

## Admin user
Screenshot from 2023-06-22 09-34-03

## Normal user
Screenshot from 2023-06-20 12-23-25

### DIFF
--- a/src/components/OAuthConnectButton.vue
+++ b/src/components/OAuthConnectButton.vue
@@ -15,6 +15,7 @@ import OpenInNewIcon from 'vue-material-design-icons/OpenInNew.vue'
 import { generateUrl } from '@nextcloud/router'
 import { showError } from '@nextcloud/dialogs'
 import { getCurrentUser } from '@nextcloud/auth'
+import { translate as t } from '@nextcloud/l10n'
 import '@nextcloud/dialogs/styles/toast.scss'
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
 import dompurify from 'dompurify'
@@ -42,8 +43,7 @@ export default {
 
 	computed: {
 		adminConfigNotOkMessage() {
-			const user = getCurrentUser()
-			if (user.isAdmin) {
+			if (getCurrentUser().isAdmin) {
 				const linkText = t('integration_openproject', 'Administration Settings > OpenProject')
 				const url = generateUrl('/settings/admin/openproject')
 				const htmlLink = `<a class="link" href="${url}" target="_blank" title="${linkText}">${linkText}</a>`
--- a/src/components/OAuthConnectButton.vue
+++ b/src/components/OAuthConnectButton.vue
@@ -7,18 +7,17 @@
 		</template>
 		{{ t('integration_openproject', 'Connect to OpenProject') }}
 	</NcButton>
-	<div v-else class="oauth-connect--message">
-		{{ adminConfigNotOkMessage }}
-	</div>
+	<div v-else class="oauth-connect--message" v-html="adminConfigNotOkMessage" /> <!-- eslint-disable-line vue/no-v-html -->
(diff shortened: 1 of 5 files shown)


src/components/OAuthConnectButton.vue
```diff
@@ -43,6 +43,14 @@ export default {
...
 	computed: {
 		adminConfigNotOkMessage() {
+			if (getCurrentUser().isAdmin) {
+				const linkText = t('integration_openproject', 'Administration Settings > OpenProject')
+				const url = generateUrl('/settings/admin/openproject')
+				const htmlLink = `<a class="link" href="${url}" target="_blank" title="${linkText}">${linkText}</a>`
+				const hintText = t('integration_openproject', 'Some OpenProject integration application settings are not working. '
+				+ 'Click here {htmlLink} to set them up properly.', { htmlLink }, null, { escape: false, sanitize: false })
+				return dompurify.sanitize(hintText, { ADD_ATTR: ['target'] })
+			}
```
SwikritiT, line 53: I've added a message like this for now. Need suggestion if this message makes sense or if we can change it so something else

SagarGi (approved):
I Checked it locally LGTM :+1:

src/components/OAuthConnectButton.vue
SagarGi: I think it fine.

tests/jest/components/__snapshots__/OAuthConnectButton.spec.js.snap
```diff
@@ -1,7 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`OAuthConnectButton.vue when the admin config is not okay should show message 1`] = `
-<div class="oauth-connect--message">
-  Some OpenProject integration application settings are not working. Please contact your Nextcloud administrator.
-</div>
-`;
+exports[`OAuthConnectButton.vue when the admin config is not okay should show message for admin user 1`] = `<div class="oauth-connect--message">Some OpenProject integration application settings are not working. Please contact your Nextcloud administrator.</div>`;
+
+exports[`OAuthConnectButton.vue when the admin config is not okay should show message for normal user 1`] = `<div class="oauth-connect--message">Some OpenProject integration application settings are not working. Please contact your Nextcloud administrator.</div>`;
```
SwikritiT, line 5: just noticed that these snap shots aren't correct

src/components/OAuthConnectButton.vue
```diff
@@ -43,6 +43,14 @@ export default {
 
 	computed: {
 		adminConfigNotOkMessage() {
+			if (getCurrentUser().isAdmin) {
+				const linkText = t('integration_openproject', 'Administration Settings > OpenProject')
+				const url = generateUrl('/settings/admin/openproject')
+				const htmlLink = `<a class="link" href="${url}" target="_blank" title="${linkText}">${linkText}</a>`
+				const hintText = t('integration_openproject', 'Some OpenProject integration application settings are not working. '
+				+ 'Click here {htmlLink} to set them up properly.', { htmlLink }, null, { escape: false, sanitize: false })
```
individual-it, line 51: Maybe something like that:

```suggestion
				const hintText = t('integration_openproject', 'Some OpenProject integration application settings are not working. '
				+ 'Configure the OpenProject integration in: {htmlLink}.', { htmlLink }, null, { escape: false, sanitize: false })
```

individual-it (approved):

PR #430, opened 2023-06-30
[OP#48842] Add new parameter in unit test client

SwikritiT:
[OP#48842] : https://community.openproject.org/projects/nextcloud-integration/work_packages/details/48842

A new parameter got added in the constructor of  `client` in nextcloud master so unit tests were failing so this PR fixes that

### DIFF
--- a/tests/lib/Service/OpenProjectAPIServiceTest.php
+++ b/tests/lib/Service/OpenProjectAPIServiceTest.php
@@ -232,8 +232,40 @@ class OpenProjectAPIServiceTest extends TestCase {
 		$client = new GuzzleClient();
 		$clientConfigMock = $this->getMockBuilder(IConfig::class)->getMock();
 
-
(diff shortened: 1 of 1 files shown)


SagarGi (approved):
LGTM :+1:  If CI is happy

SwikritiT:
Now we have another problem that the phpstan is complaining https://github.com/nextcloud/integration_openproject/actions/runs/5440325820/jobs/9893142778?pr=430

Seems like the the type of `data` was change to `psalm-type` `DataResponseType` recently

https://github.com/nextcloud/server/blob/5d9d37e2c5d928288220d17160e9b6e29f064eb8/lib/public/AppFramework/Http/DataResponse.php#L34

Now the phpstan is complaining because it doen't understand this type and is treating it as `class` which it can't find 😢

SwikritiT:
superseded by https://github.com/nextcloud/integration_openproject/pull/432

PR #431, opened 2023-06-30
[OP#48744] Show proper error message when the openproject server is not reachable

SwikritiT:
[OP#48744] : https://community.openproject.org/projects/nextcloud-integration/work_packages/48744

### Project tab
Screenshot from 2023-06-26 17-17-23

### Dashboard
Screenshot from 2023-06-26 17-17-04

### Project tab
Screenshot from 2023-06-30 15-02-32

### Dashboard
Screenshot from 2023-06-30 15-02-48

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -208,7 +208,7 @@ export default {
 				}
 			}).catch((error) => {
 				clearInterval(this.loop)
-				if (error.response && error.response.status === 404) {
+				if (error.response && (error.response.status === 404 || error.response.status === 503)) {
 					this.state = STATE.CONNECTION_ERROR
 				} else if (error.response && error.response.status === 401) {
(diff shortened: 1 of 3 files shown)


SwikritiT:
The personal setting is still like this, I don't know if we should show the same error message?
Screenshot from 2023-06-30 15-03-08

src/views/Dashboard.vue
```diff
@@ -208,7 +208,7 @@ export default {
 				}
 			}).catch((error) => {
 				clearInterval(this.loop)
-				if (error.response && error.response.status === 404) {
+				if (error.response && (error.response.status === 404 || error.response.status === 503)) {
```
individual-it, line 211: why 503? In what case do we get 503?

SwikritiT: when the openproject server is not up and we send an api request

individual-it (approved):

PR #432, opened 2023-07-03
[OP#48849] Use psalm as static code analysis tool for nextcloud master / fix phpunit tests / fix op user can be disabled

SwikritiT:
Related work packages:
1. https://community.openproject.org/projects/nextcloud-integration/work_packages/48849
2. https://community.openproject.org/projects/nextcloud-integration/work_packages/48842
3. https://community.openproject.org/projects/nextcloud-integration/work_packages/48861

## Description
**1.** The type of `data`  that we pass to `DataResponse()` was changed to `psalm-type` `DataResponseType` recently

https://github.com/nextcloud/server/blob/5d9d37e2c5d928288220d17160e9b6e29f064eb8/lib/public/AppFramework/Http/DataResponse.php#L34

The `phpstan` was complaining because it doesn't understand this type and is treating it as `class` which it can't find. In order to fix this issue this PR uses `pslam` as static code analysis tool for `master` branch of `nextcloud` 

> Note that this change was only made in `master` of nextcloud so it doesn't affect other versions. Thus we use `phpstan` for other versions.

Some of the known errors are also suppressed in `pslam` like for when `activity` app or `groupfolders` app is not available. Some changes are made in the return type in order to make both `pslam` and `phpstan` happy.

**2.** Changes of this PR https://github.com/nextcloud/integration_openproject/pull/430 is also cherrypicked here because the CI won't be green without it. Basically, a new parameter got added in the constructor of client in nextcloud master so unit tests were failing so this PR fixes that

**3.** Another error is that the `OpenProject` user gets disabled even when there's an error message. This resulted into api test failing because we test there that the user cannot be disabled. This also only started happening recently and only reproducible in the master was fixed here because CI won't be green without it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,8 +75,14 @@ jobs:
           cd server && git submodule update --init
           ./occ maintenance:install --admin-pass=admin
 
-      - name: PHP stan
-        run: make phpstan
+      - name: PHP code analysis
+        run: |
+          if [[ ${{ matrix.nextcloudVersion }} == "master" ]]
+          then
+            make psalm
+          else
+            make phpstan
+          fi
 
       - name: PHP code style
         run: composer run cs:check || ( echo 'Please run `composer run cs:fix` to format your code' && exit 1 )
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,8 @@
 		"phpstan/phpstan-phpunit": "^1.0",
 		"phpstan/extension-installer": "^1.1",
 		"behat/behat": "^3.10",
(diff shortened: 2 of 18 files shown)


SwikritiT:
We have another problem now that the `OpenProject` user get's disabled even when there's an error message. This only happens with nextcloud master. Maybe some new changes got added in the even that we are using to prevent disabling of the user.
I've created a new work package for this.
https://community.openproject.org/projects/nextcloud-integration/work_packages/48861

individual-it (approved):

SagarGi (approved):
LGTM :+1:

PR #434, opened 2023-07-10
[OP#49038] Refactor the name of the project to follow upper camel case

SwikritiT:
Part of: https://community.openproject.org/projects/nextcloud-integration/work_packages/49038

As `OpenProject Integration` is the name of the app co it should start with upper case. There were places where this naming pattern wasn't followed so this PR fixes that.


SagarGi (approved):
LGTM :+1:

PR #435, opened 2023-07-10
[OP#48843] Use the empty content component of this repo 

SwikritiT:
Related work package [OP#48843] - https://community.openproject.org/projects/nextcloud-integration/work_packages/48843

We have our own `EmptyContent` component but in the dashboard, we were using Nextcloud's one. So in this PR, I made the necessary changes to use our `EmptyContent` component everywhere.  

We were also throwing `400` for when the user didn't have the `access token` to OpenProject but that's not the right status code in this case so this PR changes the status code to `401`

### DIFF
--- a/src/components/tab/EmptyContent.vue
+++ b/src/components/tab/EmptyContent.vue
@@ -2,7 +2,8 @@
 	<div class="empty-content">
 		<div class="empty-content--wrapper">
 			<div class="empty-content--icon">
-				<LinkPlusIcon v-if="!!isAdminConfigOk && isStateOk" :size="60" />
+				<CheckIcon v-if="isStateOk && dashboard" :size="60" />
+				<LinkPlusIcon v-else-if="!!isAdminConfigOk && isStateOk && !dashboard" :size="60" />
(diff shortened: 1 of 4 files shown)


src/components/tab/EmptyContent.vue
```diff
@@ -72,12 +78,15 @@ export default {
 			} else if (this.state === STATE.FAILED_FETCHING_WORKPACKAGES) {
 				return t('integration_openproject', 'Could not fetch work packages from OpenProject')
 			} else if (this.isStateOk) {
+				if (this.dashboard) {
+					return t('integration_openproject', 'No OpenProject notifications!')
```
SagarGi, line 82:
```suggestion
					return t('integration_openproject', 'No OpenProject notifications !')
```

SagarGi (approved):
Just a small change other LGTM :+1:

src/components/tab/EmptyContent.vue
SwikritiT: I don't think we are supposed to put white space before `!` according to English punctuation.

PR #438, opened 2023-07-14
[#49133] throw better error message when admin tries to delete/disable OP user/group

SwikritiT:
Related workpackage[OP#49133]: https://community.openproject.org/work_packages/49133

We are protecting user/group Openproject from being deleted/disabled but the error message displayed is `An error occurred during the request. Unable to proceed` which is vague. So show proper error messages so that it's known that this user is protected.

In this PR, we let the user know that the OP user/group is protected and redirect them towards open project troubleshooting docs for further information

The error message displayed now is

Screenshot from 2023-07-21 12-47-45

### DIFF
--- a/lib/Listener/BeforeGroupDeletedListener.php
+++ b/lib/Listener/BeforeGroupDeletedListener.php
@@ -59,8 +59,10 @@ class BeforeGroupDeletedListener implements IEventListener {
 
 		$group = $event->getGroup();
 		if ($group->getGID() === Application::OPEN_PROJECT_ENTITIES_NAME) {
-			$this->logger->info('Group openproject cannot be deleted');
-			throw new OCSBadRequestException();
+			$this->logger->error('Group "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted. Please check the documentation "https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" for further information.');
+			throw new OCSBadRequestException('<p>&nbsp;Group "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted.
+			Please check the <a style="color:var(--color-primary-default)" href="https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting"
+			target="_blank"><u>troubleshooting guide</u></a> for further information.</p>');
 		}
 	}
 }
--- a/lib/Listener/BeforeUserDeletedListener.php
+++ b/lib/Listener/BeforeUserDeletedListener.php
@@ -57,8 +57,10 @@ class BeforeUserDeletedListener implements IEventListener {
 		}
 		$user = $event->getUser();
 		if ($user->getUID() === Application::OPEN_PROJECT_ENTITIES_NAME) {
-			$this->logger->info('User openproject cannot be deleted');
-			throw new OCSBadRequestException();
+			$this->logger->error('User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted. Please check the documentation "https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" for further information.');
+			throw new OCSBadRequestException('<p>&nbsp;User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted.
+			Please check the <a style="color:var(--color-primary-default)" href="https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting"
+			target="_blank"><u>troubleshooting guide</u></a> for further information.</p>');
 		}
 	}
 }
--- a/lib/Listener/UserChangedListener.php
+++ b/lib/Listener/UserChangedListener.php
@@ -63,9 +63,12 @@ class UserChangedListener implements IEventListener {
 		if ($event->getUser()->getUID() === Application::OPEN_PROJECT_ENTITIES_NAME) {
 			$feature = $event->getFeature();
 			if ($feature === 'enabled' && !$event->getValue()) {
-				$this->logger->info('User openproject cannot be disabled');
+				$this->logger->error('User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be disabled. Please check the documentation "https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" for further information.');
 				$event->getUser()->setEnabled();
-				throw new OCSBadRequestException();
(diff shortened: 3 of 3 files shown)


lib/Listener/BeforeGroupDeletedListener.php
```diff
@@ -59,8 +59,8 @@ public function handle(Event $event): void {
 
 		$group = $event->getGroup();
 		if ($group->getGID() === Application::OPEN_PROJECT_ENTITIES_NAME) {
-			$this->logger->info('Group openproject cannot be deleted');
-			throw new OCSBadRequestException();
+			$this->logger->info('Group "OpenProject" is protected by OpenProject Integration, thus cannot be deleted.');
```
wielinde, line 62:
```suggestion
			$this->logger->info('Group "OpenProject" is needed protected by the app "OpenProject Integration", thus cannot be deleted. If you really need to delete the group, first deactivate the app. Then you can delete the group. Deactivating and reactivating the app is safe. No app data will be deleted.');
```

SwikritiT: > @SwikritiT Does it make sense to add more info/assistance to the error message?

Yes, a detailed message might be more beneficial. I'll update the message


> Also, isn't there a similar piece of code protecting the deletion of the user? At least that was what I tried and got stuck.

True, we throw a similar message when the `OpenProject` user is tried to be disabled/deleted. I have updated those messages too in the same PR

SwikritiT: @wielinde displaying the long message in the UI is not flattering.  It looks something like this, and I don't think the styling can be changed from our app.

Screenshot from 2023-07-17 10-09-41


I've changed the message to something like this.

Screenshot from 2023-07-17 11-28-34



Redirecting the user towards the log where the details about how the user can be deleted is wrriten. Will that be okay?

SwikritiT: If I use break then the message looks like this 

Screenshot from 2023-07-17 12-10-30



> Note: that the first line is from nextcloud.

lib/Listener/BeforeUserDeletedListener.php
```diff
@@ -57,8 +57,8 @@ public function handle(Event $event): void {
 		}
 		$user = $event->getUser();
 		if ($user->getUID() === Application::OPEN_PROJECT_ENTITIES_NAME) {
-			$this->logger->info('User openproject cannot be deleted');
-			throw new OCSBadRequestException();
+			$this->logger->error('User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted. Please check the documentation "https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" for further information.');
+			throw new OCSBadRequestException('<p>User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted. Please check the <a style="color:var(--color-primary-default)" href="https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" target="_blank"><u>documentation</u></a> for further information.</p>');
```
SagarGi, line 61:
```suggestion
			throw new OCSBadRequestException('<p> User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted. Please check the <a style="color:var(--color-primary-default)" href="https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" target="_blank"><u>documentation</u></a> for further information.</p>');
```

lib/Listener/UserChangedListener.php
```diff
@@ -63,9 +63,9 @@ public function handle(Event $event): void {
 		if ($event->getUser()->getUID() === Application::OPEN_PROJECT_ENTITIES_NAME) {
 			$feature = $event->getFeature();
 			if ($feature === 'enabled' && !$event->getValue()) {
-				$this->logger->info('User openproject cannot be disabled');
+				$this->logger->error('User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be disabled. Please check the documentation "https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" for further information.');
 				$event->getUser()->setEnabled();
-				throw new OCSBadRequestException();
+				throw new OCSBadRequestException('<p>User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be disabled. Please check the <a style="color:var(--color-primary-default)" href="https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" target="_blank"><u>documentation</u></a> for further information.</p>');
```
SagarGi, line 68:
```suggestion
				throw new OCSBadRequestException('<p> User "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be disabled. Please check the <a style="color:var(--color-primary-default)" href="https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" target="_blank"><u>documentation</u></a> for further information.</p>');
```

lib/Listener/BeforeGroupDeletedListener.php
```diff
@@ -59,8 +59,8 @@ public function handle(Event $event): void {
 
 		$group = $event->getGroup();
 		if ($group->getGID() === Application::OPEN_PROJECT_ENTITIES_NAME) {
-			$this->logger->info('Group openproject cannot be deleted');
-			throw new OCSBadRequestException();
+			$this->logger->error('Group "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted. Please check the documentation "https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" for further information.');
+			throw new OCSBadRequestException('<p> Group "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted. Please check the <a style="color:var(--color-primary-default)" href="https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" target="_blank"><u>documentation</u></a> for further information.</p>');
```
SagarGi, line 63:
```suggestion
			throw new OCSBadRequestException('<p>&nbsp;Group "OpenProject" is needed to be protected by the app "OpenProject Integration", thus cannot be deleted. Please check the <a style="color:var(--color-primary-default)" href="https://www.openproject.org/docs/system-admin-guide/integrations/nextcloud/#troubleshooting" target="_blank"><u>documentation</u></a> for further information.</p>');
```

lib/Listener/UserChangedListener.php
SagarGi, line 68: instead of `documentation` can we say `troubleshooting guide`?

SwikritiT: yes sure

SagarGi (approved):
LGTM :+1:

PR #439, opened 2023-07-17
[OP#48106] Enable links preview even when the global search for workpackages is not enabled

SwikritiT:
Related workpackage: https://community.openproject.org/projects/nextcloud-integration/work_packages/48106/activity

Previously, the link previews were only displayed when the user `enables the global search for workpackages` settings. We decided that the links previews should be enabled by default so this PR removes that codependency.

## Demo

https://github.com/nextcloud/integration_openproject/assets/41103328/62f2dc36-7b14-4ecc-ab3f-527bbd116d67


SagarGi (commented):
LGTM :+1:

SagarGi (approved):

PR #443, opened 2023-07-25
[OP#49381] drop support for nc 23

SwikritiT:
Related work package: https://community.openproject.org/projects/nextcloud-integration/work_packages/49381

As a general rule, Nextcloud Enterprise always supports the last three major releases. That might be longer than the non-entprise releases. As nextcloud 26 is released we can drop support for nextcloud 23


SagarGi (approved):
LGTM :+1:

PR #444, opened 2023-07-26
[OP#40971]support multiple files to a work-package

SagarGi:
### Descriptions
This PR adds a feature for multiples files to link to a work package.
- Adds implementation for multiple files to link to a work package
- Adds php unit tests
- Adds UI test 

### Related WorkPackage:
https://community.openproject.org/projects/nextcloud-integration/work_packages/40971/activity


lib/Service/OpenProjectAPIService.php
```diff
@@ -704,53 +704,61 @@ public function getNode($userId, $fileId) {
...
-		int $fileId,
-		string $fileName,
+		array $values,
 		string $userId
-	): int {
-		$file = $this->getNode($userId, $fileId);
-		if (!$file->isReadable()) {
-			throw new NotPermittedException();
+	): array {
+		$fileIfnos = $values['fileinfo'];
```
individual-it, line 723:
```suggestion
		$fileInfos = $values['fileinfo'];
```

src/components/tab/SearchInput.vue
```diff
@@ -141,22 +156,28 @@ export default {
...
-					'Content-Type': 'application/x-www-form-urlencoded',
+			// since we can link multiple files now we send file information required in array (whether its only one value or multiple)
+			let fileInfoForBody = []
+			let successMessage
+			if (this.isSearchFrom === SEARCHFROM.PROJECT_TAB) {
+				fileInfoForBody.push(this.fileInfo)
+				successMessage = 'Work package linked successfully!'
+			} else if (this.isSearchFrom === SEARCHFROM.LINK_MULTIPLE_MODAL) {
+				fileInfoForBody = this.fileInfo
+				successMessage = 'Work package linked successfully for selected files!'
```
individual-it, line 167: translation missing

src/components/tab/SearchInput.vue
```diff
@@ -141,22 +156,28 @@ export default {
...
-			params.append('fileName', this.fileInfo.name)
-			const config = {
-				headers: {
-					'Content-Type': 'application/x-www-form-urlencoded',
+			// since we can link multiple files now we send file information required in array (whether its only one value or multiple)
+			let fileInfoForBody = []
+			let successMessage
+			if (this.isSearchFrom === SEARCHFROM.PROJECT_TAB) {
+				fileInfoForBody.push(this.fileInfo)
+				successMessage = 'Work package linked successfully!'
```
individual-it, line 164: translation missing

src/components/tab/SearchInput.vue
```diff
@@ -183,7 +204,7 @@ export default {
 			if (response.status === 200) await this.processWorkPackages(response.data, fileId)
 			if (this.isStateLoading) this.state = STATE.OK
 		},
-		async processWorkPackages(workPackages, fileId) {
+		async processWorkPackages(workPackages, fileId = null) {
```
individual-it, line 207: in which case would fileId be null?

src/components/tab/SearchInput.vue
```diff
@@ -192,7 +213,10 @@ export default {
 							this.searchResults.push(workPackage)
 
 						} else {
-							workPackage.fileId = fileId
+							if (this.isSearchFrom === SEARCHFROM.PROJECT_TAB) {
```
individual-it, line 216: IMO this check should be done in the function that calls `processWorkPackages`

src/components/tab/SearchInput.vue
```diff
@@ -192,7 +213,10 @@ export default {
...
+								// for now multiple link of files to a workpackage, fileId is not sent for filtration process
```
individual-it, line 217: so why then even call `getAdditionalMetaData` at all?

src/views/LinkMultipleFilesModal.vue
```diff
@@ -0,0 +1,150 @@
...
+			} else {
+				this.state = STATE.ERROR
+			}
+		},
+		closeRequestModal() {
+			this.fileInfos = []
+			this.alreadyLinkedWorkPackage = []
+			this.show = false
+		},
+		async fetchWorkpackagesForMultipleFileLink(fileId) {
```
individual-it, line 90: I thought we don't have to check if a file is already linked

webpack.js
```diff
@@ -19,6 +19,7 @@ webpackConfig.entry = {
 	dashboard: { import: path.join(__dirname, 'src', 'dashboard.js'), filename: appId + '-dashboard.js' },
 	'openproject-tab': { import: path.join(__dirname, 'src', 'projectTab.js'), filename: appId + '-projectTab.js' },
 	fileActions: { import: path.join(__dirname, 'src', 'fileActions.js'), filename: appId + '-fileActions.js' },
+	filesplugin: { import: path.join(__dirname, 'src', 'filesplugin.js'), filename: appId + '-filesplugin.js' },
```
individual-it, line 22:
```suggestion
	filesPlugin: { import: path.join(__dirname, 'src', 'filesPlugin.js'), filename: appId + '-filesPlugin.js' },
```

lib/Service/OpenProjectAPIService.php
```diff
@@ -704,53 +704,61 @@ public function getNode($userId, $fileId) {
...
 	 * @throws OpenprojectResponseException
-	 * @return int
+	 *
+	 * @return array<int>
 	 */
 	public function linkWorkPackageToFile(
-		int $workpackageId,
-		int $fileId,
-		string $fileName,
+		array $values,
```
individual-it, line 720: this will be so confusing to know what "values" needed to be send here, same in `linkWorkPackageToFile`
we should 
- well document what `$values` expects
- maybe not expecting an array but an object, that will reduce the chances for errors
- check the content of the passed object/array (my guess is, that if we now don't supply the correct values in `$values` the whole function will fall over with `undefined index`)
- have tests, that check if something is wrong with the given parameters

individual-it (changes requested):

src/components/tab/SearchInput.vue
SagarGi: will be translated in this function `showSuccess(
					t('integration_openproject', successMessage)
				)`

src/components/tab/SearchInput.vue
SagarGi: in case we search from the multiple link file modal or also from the link picker we do not need the filter the search result by id. in project tab it was done because of some inconsistency in the UI. But is different in this case.

src/views/LinkMultipleFilesModal.vue
SagarGi: we check for only single file if the file is already linked to a work package now but not all.

src/components/tab/SearchInput.vue
individual-it: That does not work, because the translation tool does not run the code and so it has no idea about the content of `successMessage`. You have to use `t()` where ever the literal string is supplied

SagarGi: oh ok sure.

lib/Service/OpenProjectAPIService.php
SagarGi: Added unit test with validation.

src/views/LinkMultipleFilesModal.vue
SagarGi: For project tab fetch there are some extra checks that we do not require in the fetch from the link modal. So The function is a bit different. May be its better to have separate function for both.

lib/Controller/OpenProjectAPIController.php
```diff
@@ -197,15 +198,17 @@ public function linkWorkPackageToFile(int $workpackageId, int $fileId, string $f
...
 		try {
 			$result = $this->openprojectAPIService->linkWorkPackageToFile(
-				$workpackageId,
-				$fileId,
-				$fileName,
+				$values,
 				$this->userId,
 			);
 		} catch (OpenprojectErrorException $e) {
 			return new DataResponse($e->getMessage(), Http::STATUS_BAD_REQUEST);
```
SwikritiT, line 205: Maybe we can do it like this since we're throwing the same status code?
```suggestion
		} catch (OpenprojectErrorException | InvalidArgumentException $e) {
			return new DataResponse($e->getMessage(), Http::STATUS_BAD_REQUEST);
```

lib/Controller/OpenProjectAPIController.php
```diff
@@ -197,15 +198,17 @@ public function linkWorkPackageToFile(int $workpackageId, int $fileId, string $f
...
 		} catch (NotPermittedException | NotFoundException $e) {
 			return new DataResponse('file not found', Http::STATUS_NOT_FOUND);
+		} catch (InvalidArgumentException $e) {
+			return new DataResponse([
+				"error" => $e->getMessage()
+			], Http::STATUS_BAD_REQUEST);
 		} catch (Exception $e) {
```
SwikritiT, line 212: While you're at it can you refactor this to something like this? seems like we missed the `\`
```suggestion
		} catch (\Exception $e) {
```

lib/Service/OpenProjectAPIService.php
```diff
@@ -704,53 +705,82 @@ public function getNode($userId, $fileId) {
...
+		$fileInfos = $values['fileinfo'];
+		$elements = [];
+		// multiple files can also be linked to a single work package
+		foreach ($fileInfos as $fileInfo) {
+			if (!isset($fileInfo['id']) || !isset($fileInfo['name'])) {
+				throw new InvalidArgumentException('invalid data');
+			}
+			if (!is_int($fileInfo['id']) || $fileInfo['name'] === '' || !is_string($fileInfo['name'])) {
+				throw new InvalidArgumentException('invalid data');
+			}
```
SwikritiT, line 748: can't we calculate them in a single `if` statement and should these be `&&` operations?
```suggestion
			if (!isset($fileInfo['id']) && !isset($fileInfo['name']) !is_int($fileInfo['id']) && $fileInfo['name'] === '' |&& !is_string($fileInfo['name'])) {
				throw new InvalidArgumentException('invalid data');
			}
```

src/components/tab/EmptyContent.vue
```diff
@@ -63,6 +63,10 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		isLinkMultipleModal: {
```
SwikritiT, line 66: How about this?
```suggestion
		isMultipleWorkpackageLinking: {
```

src/components/tab/SearchInput.vue
```diff
@@ -126,7 +137,12 @@ export default {
 		},
 		async asyncFind(query) {
 			this.resetState()
-			await this.debounceMakeSearchRequest(query, this.fileInfo.id, this.isSmartPicker)
+			if (this.isSearchFrom === SEARCHFROM.PROJECT_TAB) {
+				await this.debounceMakeSearchRequest(query, this.fileInfo.id)
+			} else {
+				// we do not need to provide file id incase of searching through link multiple files to workpakcage and through linkpicker
```
SwikritiT, line 143:
```suggestion
				// we do not need to provide a file id incase of searching through link multiple files to work package and through smart picker
```

src/components/tab/SearchInput.vue
```diff
@@ -141,33 +157,38 @@ export default {
...
 				return
 			}
-			const params = new URLSearchParams()
-			params.append('workpackageId', selectedOption.id)
-			params.append('fileId', this.fileInfo.id)
-			params.append('fileName', this.fileInfo.name)
-			const config = {
-				headers: {
-					'Content-Type': 'application/x-www-form-urlencoded',
+			// since we can link multiple files now we send file information required in array (whether its only one value or multiple)
```
SwikritiT, line 160:
```suggestion
			// since we can link multiple files now we send file information required in an array (whether it's only one value or multiple)
```

src/components/tab/SearchInput.vue
```diff
@@ -141,33 +157,38 @@ export default {
...
+			let fileInfoForBody = []
+			let successMessage
+			if (this.isSearchFrom === SEARCHFROM.PROJECT_TAB) {
+				fileInfoForBody.push(this.fileInfo)
+				successMessage = t('integration_openproject', 'Work package linked successfully!')
```
SwikritiT, line 165: I see in Figma the toast message is something like this, maybe we didn't follow through the message the previous time. Can we change it, this seems more accurate than what we are displaying right now. 
```suggestion
				successMessage = t('integration_openproject', 'Link to work package created successfully!')
```

src/components/tab/SearchInput.vue
```diff
@@ -141,33 +157,38 @@ export default {
...
+			} else if (this.isSearchFrom === SEARCHFROM.LINK_MULTIPLE_MODAL) {
+				fileInfoForBody = this.fileInfo
+				successMessage = t('integration_openproject', 'Work package linked successfully for selected files!')
```
SwikritiT, line 168:
```suggestion
				successMessage = t('integration_openproject', 'Links to work package created successfully for selected files!')
```

src/filesPlugin.js
```diff
@@ -0,0 +1,74 @@
...
+			}
+			OCA.OpenProject.LinkMultipleFilesModalVue.$children[0].setFileInfos(fileInfos)
+			OCA.OpenProject.LinkMultipleFilesModalVue.$children[0].showModal()
+		},
+	}
+})()
+
+OC.Plugins.register('OCA.Files.FileList', OCA.OpenProject.FilesPlugin)
+
+const modalId = 'exampleModal'
```
SwikritiT, line 64: maybe something like this?
```suggestion
const modalId = 'multipleFileLinkModal'
```

src/utils.js
```diff
@@ -48,3 +48,8 @@ export const FORM = {
 	GROUP_FOLDER: 3,
 	APP_PASSWORD: 4,
 }
+
+export const SEARCHFROM = {
+	PROJECT_TAB: 'project-tab',
+	LINK_MULTIPLE_MODAL: 'link-multiple-modal',
```
SwikritiT, line 54: Maybe something like this? Right now it seems like we're linking multiple modals
```suggestion
	LINK_MULTIPLE_FILES_MODAL: 'link-multiple-files-modal',
```

src/utils.js
```diff
@@ -48,3 +48,8 @@ export const FORM = {
 	GROUP_FOLDER: 3,
 	APP_PASSWORD: 4,
 }
+
+export const SEARCHFROM = {
```
SwikritiT, line 52: How about something like this?
```suggestion
export const SEARCH_WORKPACKAGES_FROM = {
```

src/views/LinkMultipleFilesModal.vue
```diff
@@ -0,0 +1,147 @@
...
+		LoadingIcon,
+	},
+	data() {
+		return {
+			show: false,
+			state: STATE.LOADING,
+			fileInfos: [],
+			alreadyLinkedWorkPackage: [],
+			isAdminConfigOk: loadState('integration_openproject', 'admin-config-status'),
+			isSearchFrom: SEARCHFROM.LINK_MULTIPLE_MODAL,
```
SwikritiT, line 54:
```suggestion
			isSearchWorkpackageFrom: SEARCHFROM.LINK_MULTIPLE_MODAL,
```

SwikritiT (commented):
some suggestions

SwikritiT:
@SagarGi I made about a `100` files 

```bash
#!/usr/bin/env bash

for (( c=1; c<=110; c++ ))
do 
    folderName="test$c"
    echo ${folderName}
    # curl -X MKCOL -u admin:admin http://localhost/nextcloud/stable26/remote.php/dav/files/admin/${folderName}
    curl -X PUT -u admin:admin http://localhost/nextcloud/stable26/remote.php/webdav/${folderName}.txt -d "hello world"
done
```

And tried to link it to a work package, but it gives me `400` and none of the files are linked. This can be a bug from `openproject` side as well can we make sure that this is not a problem on our end?

SagarGi:
> @SagarGi I made about a `100` files
> 
> ```shell
> #!/usr/bin/env bash
> 
> for (( c=1; c<=110; c++ ))
> do 
>     folderName="test$c"
>     echo ${folderName}
>     # curl -X MKCOL -u admin:admin http://localhost/nextcloud/stable26/remote.php/dav/files/admin/${folderName}
>     curl -X PUT -u admin:admin http://localhost/nextcloud/stable26/remote.php/webdav/${folderName}.txt -d "hello world"
> done
> ```
> 
> And tried to link them all to a work package, but it gives me `400`, and none of the files are linked. This can be a bug from `openproject` side as well but can we make sure if it is a problem on our end?




toMany


Its the error from the openproject with message `To Many`. May be a clear information error might be necessary in this case. Can also be done in another PR

SwikritiT:
> > @SagarGi I made about a `100` files
> > ```shell
> > #!/usr/bin/env bash
> > 
> > for (( c=1; c<=110; c++ ))
> > do 
> >     folderName="test$c"
> >     echo ${folderName}
> >     # curl -X MKCOL -u admin:admin http://localhost/nextcloud/stable26/remote.php/dav/files/admin/${folderName}
> >     curl -X PUT -u admin:admin http://localhost/nextcloud/stable26/remote.php/webdav/${folderName}.txt -d "hello world"
> > done
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And tried to link them all to a work package, but it gives me `400`, and none of the files are linked. This can be a bug from `openproject` side as well but can we make sure if it is a problem on our end?
> 
> @SwikritiT toMany
> 
> Its the error from the openproject with message `To Many`. May be a clear information error might be necessary in this case. Can also be done in another PR

the limit seems to be 20 in a single request, we can show some error message if the limit is crossed maybe even from the front end but yes can be part of other PR

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -444,14 +468,20 @@ describe('SearchInput.vue', () => {
...
+								id: 111,
+								name: 'file.txt',
+							},
+						],
+					},
+				}
 				expect(postSpy).toBeCalledWith(
 					'http://localhost/apps/integration_openproject/work-packages',
-					expectedParams,
-					{ headers: { 'Content-Type': 'application/x-www-form-urlencoded' } }
```
SwikritiT, line 454: maybe we can check `Content-type: application/json` now?

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -543,6 +573,250 @@ describe('SearchInput.vue', () => {
...
+						},
+					}
+					expect(postSpy).toBeCalledWith(
+						'http://localhost/apps/integration_openproject/work-packages',
+						body
+					)
+					postSpy.mockRestore()
+				})
+
+				it('should show an error when linking failed', async () => {
```
SwikritiT, line 636:
```suggestion
				it('should show an error when linking fails', async () => {
```

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -543,6 +573,250 @@ describe('SearchInput.vue', () => {
...
+		const multipleFileInfos = [{
+			id: 123,
+			name: 'logo.png',
+		},
+		{
+			id: 456,
+			name: 'pogo.png',
+		}]
+		describe('single file selected', () => {
+			describe('click on a workpackage option', () => {
```
SwikritiT, line 592:
```suggestion
			describe('select a work package for linking', () => {
```

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -543,6 +573,250 @@ describe('SearchInput.vue', () => {
...
+							},
+						],
+					)
+					axiosSpy.mockRestore()
+				})
+
+			})
+
+			describe('multiple files selected', () => {
+				describe('click on a workpackage option', () => {
```
SwikritiT, line 702:
```suggestion
				describe('select a work package for linking', () => {
```

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -543,6 +573,250 @@ describe('SearchInput.vue', () => {
...
+						}
+						expect(postSpy).toBeCalledWith(
+							'http://localhost/apps/integration_openproject/work-packages',
+							body
+						)
+						postSpy.mockRestore()
+					})
+
+					it('should show an error when linking failed', async () => {
+						const err = new Error()
```
SwikritiT, line 747:
```suggestion
					it('should show an error when linking fails', async () => {
						const err = new Error()
```

src/components/tab/SearchInput.vue
```diff
@@ -62,6 +62,11 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		isSearchWorkpackageFrom: {
```
individual-it, line 65:
```suggestion
		searchSource: {
```

individual-it (changes requested):

individual-it:
When there is no connection yet to OP the screen shows:
grafik

In other places we show
grafik

SagarGi:
@SwikritiT @individual-it The linking of files more than 20 in chunks can be done in another PR since this PR seems big itself. So this can be reviewed again.

SwikritiT (commented):
tested locally works for me

individual-it (approved):

SwikritiT (approved):

PR #446, opened 2023-07-27
Encrypt the oauth client secrets

SwikritiT:
## Description
The server implemented the feature to encrypt the oauth client secret in https://github.com/nextcloud/server/pull/38398 
but these changes were not adapted to our app so the OAuth connection wasn't successful from the open project side.

As this feature is only available from the nextcloud `25.0.8` there's a check to ensure the version of the server.

## Related workpackage
OP#49416 - https://community.openproject.org/projects/nextcloud-integration/work_packages/49416

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,14 +155,11 @@ jobs:
     name: API tests
     strategy:
       matrix:
-        nextcloudVersion: [ stable22, stable23, stable24, stable25, stable26, stable27 ]
+        nextcloudVersion: [ stable23, stable24, stable25, stable26, stable27 ]
         phpVersionMajor: [ 7, 8 ]
         phpVersionMinor: [ 4, 1 ]
         database: [pgsql, mysql]
         exclude:
-          - nextcloudVersion: stable22
-            phpVersionMajor: 8
-            phpVersionMinor: 1
           - nextcloudVersion: stable23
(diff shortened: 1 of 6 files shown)


SwikritiT:
The api test realted to the files id endpoint keep failing with stable22 because for some reason activity app is not being downloaded https://github.com/nextcloud/integration_openproject/actions/runs/5690981026/job/15425312895#step:16:157
However, these changes should not affect the working of the files id endpoint.


In this run https://github.com/nextcloud/integration_openproject/actions/runs/5691122946  I removed the stable22 for api tests to be sure that tests are passing for other stable versions. If the CI is green for that one I'll bring back the stable22 and if the CI is red we might need to merge with red CI, as I don't see any other solution to fix the activity app not being enabled for stable 22

SagarGi (approved):
LGTM :+1:

SwikritiT:
This is the run where all the unit test passed https://github.com/nextcloud/integration_openproject/actions/runs/5689688522 

This is the run where all the api tests passed https://github.com/nextcloud/integration_openproject/actions/runs/5688758938

The do pass

SwikritiT:
The CI errors are unrelated to this PR, it's because the activity app is not installed for stable 22, it gets downloaded sometimes other times not. I cannot make all the pipelines pass in a single run but they do pass. I'm merging this PR to unblock the release

PR #448, opened 2023-08-01
[OP#48106] use same component as project tab for searching in smart picker

SwikritiT:
Previously we used the default search provider for the smart picker but we this PR we define the custom smart picker so that the search is similar to the one we have implemented in the project tab.

## Demonstration

https://github.com/nextcloud/integration_openproject/assets/41103328/b8e2a927-1728-4b45-a84d-fcee513501f0

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -196,9 +196,8 @@ export default {
 					if (this.isStateLoading) {
 						if (this.isSmartPicker) {
 						   workPackage = await workpackageHelper.getAdditionalMetaData(workPackage)
-						   if (this.isStateLoading) {
-							   this.searchResults.push(workPackage)
-							}
+							this.searchResults.push(workPackage)
+
 						} else {
 							workPackage.fileId = fileId
 							workPackage = await workpackageHelper.getAdditionalMetaData(workPackage)
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -132,13 +132,9 @@ export default {
 		},
 		async asyncFind(query) {
 			this.resetState()
-			if (this.isSmartPicker) {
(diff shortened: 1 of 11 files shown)


src/components/tab/SearchInput.vue
```diff
@@ -170,13 +194,20 @@ export default {
...
-						workPackage.fileId = fileId
-						workPackage = await workpackageHelper.getAdditionalMetaData(workPackage)
-						const alreadyLinked = this.linkedWorkPackages.some(el => el.id === workPackage.id)
-						const alreadyInSearchResults = this.searchResults.some(el => el.id === workPackage.id)
-						// check the state again, it might have changed in between
-						if (!alreadyInSearchResults && !alreadyLinked && this.isStateLoading) {
-							this.searchResults.push(workPackage)
+						if (this.isSmartPicker) {
+						   workPackage = await workpackageHelper.getAdditionalMetaData(workPackage)
+						   if (this.isStateLoading) {
```
SagarGi, line 199: is this if again necessary for state loading?

src/views/WorkPackagePickerElement.vue
```diff
@@ -0,0 +1,97 @@
...
+		linkedWorkPackages: [],
+		state: STATE.OK,
+		isAdminConfigOk: loadState('integration_openproject', 'admin-config-status'),
+	}),
+
+	computed: {
+	},
+
+	mounted() {
+	},
```
SagarGi, line 75:
```suggestion
	
```
might be added when needed.

src/components/tab/SearchInput.vue
SwikritiT: no not needed thanks for pointing it out

SagarGi (approved):
LGTM :+1:

PR #452, opened 2023-08-07
[OP#46179] Fix managed project folders label

SwikritiT:
Part of: https://community.openproject.org/projects/nextcloud-integration/work_packages/46179

The label was "Automatic managed folders" but it should be "Automatically managed folder" so this PR fixes that

## Before
Screenshot from 2023-08-07 14-24-07

## After 
Screenshot from 2023-08-07 14-33-51


SagarGi (approved):
LGTM :+1:

PR #453, opened 2023-08-08
[OP#49182] Fix the form heading index colour in different themes

SwikritiT:
Related workpackage[OP#49182]: https://community.openproject.org/projects/nextcloud-integration/work_packages/49182

Our previous fix for fixing the colours of labels in different themes didn't work if the nextcloud theme was selected to be system default theme and the system theme was dark theme. This PR fixes that.

### system default theme with system having dark theme

Screenshot from 2023-08-08 15-27-20

### system default theme with system having dark theme

Screenshot from 2023-08-08 14-58-24

### dark theme 
Screenshot from 2023-08-08 14-58-53

### dark theme high contrast 

Screenshot from 2023-08-08 14-59-04

### light theme
Screenshot from 2023-08-08 14-58-42

### light theme high contrast
Screenshot from 2023-08-08 14-59-18

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -4,7 +4,9 @@
 		<div class="openproject-server-host">
 			<FormHeading index="1"
 				:title="t('integration_openproject', 'OpenProject server')"
-				:is-complete="isServerHostFormComplete" />
+				:is-complete="isServerHostFormComplete"
+				:is-dark-theme="isDarkTheme" />
+
 			<FieldValue v-if="isServerHostFormInView"
 				is-required
 				class="pb-1"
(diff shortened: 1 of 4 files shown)


src/components/AdminSettings.vue
```diff
@@ -528,14 +534,13 @@ export default {
...
-		isDarkTheme() {
-			return !!document.querySelector("body[data-themes*='dark']")
-
-		},
 	},
 	created() {
 		this.init()
 	},
+	mounted() {
+		this.isDarkTheme = window.getComputedStyle(this.$el).getPropertyValue('--background-invert-if-dark') === 'invert(100%)'
```
individual-it, line 542: this looks like a bridle way to detect the theme

SwikritiT: true, but I couldn't find any other way to do it.

SagarGi: Yes, for the system default theme we might need to do it explicitly.

SagarGi (commented):
Checked it manually. LGTM :+1:

individual-it (approved):

PR #456, opened 2023-08-10
[OP#49621] fix app password form becomes inactive

SwikritiT:
Fixes: https://community.openproject.org/projects/nextcloud-integration/work_packages/49621/

Our previous check for `hasAppPassword` expects the user "OpenProject" to have only one app password. But in cases where the user "OpenProject" has multiple app passwords for instance if the user is logged in somewhere it will create a session app password, this check will return `false` which will result in the project folder section in the administrative form being disabled.
```php
# previous code
return sizeof($this->tokenProvider->getTokenByUser(Application::OPEN_PROJECT_ENTITIES_NAME)) === 1;
```

So in this PR we loop through all the tokens and only return true if the token with the name "OpenProject" exists and also while deleting, delete all the tokens with the name "OpenProject". This fixes the issue we are facing

### DIFF
--- a/tests/lib/Service/OpenProjectAPIServiceTest.php
+++ b/tests/lib/Service/OpenProjectAPIServiceTest.php
@@ -1755,7 +1755,16 @@ class OpenProjectAPIServiceTest extends TestCase {
 			->method('getTokenByUser')
 			->with(Application::OPEN_PROJECT_ENTITIES_NAME)
 			->willReturn([$tokenMock]);
-		$service = $this->getServiceMock([], null, null, null, null, null, null, null, null, $tokenProviderMock);
+		$service = $this->getServiceMock([],
+			null,
+			null,
+			null,
+			null,
+			null,
+			null,
+			null,
+			null,
+			$tokenProviderMock);
 		$this->assertTrue($service->hasAppPassword());
 	}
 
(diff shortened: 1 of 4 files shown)


individual-it (commented):
Is is possible to cover that with unit tests?

SagarGi:
> Is is possible to cover that with unit tests?

Also we can add API test for it as well. @SwikritiT

SagarGi (commented):
I checked it manually  and seems to be working more fine now.

SwikritiT:
@SagarGi @individual-it I added the tests can you review it again please

SagarGi (commented):
LGTM :+1:

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -1739,6 +1744,36 @@ public function testIsSystemReadyForGroupFolderSetUpUserOrGroupExistsException(
 		$service->isSystemReadyForProjectFolderSetUp();
 	}
 
+	public function testProjectFolderHasAppPassword(): void {
```
individual-it, line 1747: lets add also a test where multiple tokens are returned but only one is an OpenProject token.
and what happens if multiple tokens have the name `OpenProject`?

SwikritiT: If there are multiple OpenProject tokens then `hasAppPassword` will return `true` but while deleting it will delete all the app passwords before creating a new one. Is that okay or do we want to check if there's only one token named `OpenProject` and if there are multiple then send an error message?

individual-it (approved):

PR #466, opened 2023-08-22
[WP#49755]Focus search input for workpackage when displayed

SagarGi:
### Description
Search input from `smartpicker` was not focused when it was displayed instead the closed button was displayed. This PR makes sure that the search input is focused when user opens the search for the workpackages from any where i.e from `smart-picker`, `tab` or `multiple-files-link-modal`.

### Relates Issue:
https://community.openproject.org/work_packages/49755/activity

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -2,6 +2,7 @@
 	<div id="searchBar">
 		<NcSelect ref="workPackageSelect"
 			class="searchInput"
+			input-id="searchInput"
 			:placeholder="placeholder"
 			:options="filterSearchResultsByFileId"
 			:user-select="true"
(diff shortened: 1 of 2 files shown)


src/components/tab/SearchInput.vue
```diff
@@ -107,17 +108,10 @@ export default {
...
 			}
 		},
 	},
 	methods: {
-		emptySearchInput() {
-			// FIXME: https://github.com/shentao/vue-multiselect/issues/633
-			if (this.$refs.workPackageSelect?.$refs?.VueMultiselect?.search) {
-				this.$refs.workPackageSelect.$refs.VueMultiselect.search = ''
-			}
-		},
```
SwikritiT, line 120: This was for the multiselect but as we are using the NcSelect which is a extension of VueSelect it should be okay to remove this code.

src/components/tab/SearchInput.vue
```diff
@@ -2,6 +2,7 @@
 	<div id="searchBar">
 		<NcSelect ref="workPackageSelect"
 			class="searchInput"
+			input-id="searchInput"
```
SwikritiT, line 5: Don't think we need this here
```suggestion
			
```

SwikritiT (approved):
LGTM

individual-it (approved):

PR #468, opened 2023-08-23
[OP#49754] fix File(s)info endpoint does not return fileID for group folders

SwikritiT:
[OP#49754] - https://community.openproject.org/work_packages/49754

For groupfolders if the fileId's are sent as string then the request returned `404` and also for group folders the internal path is `__groupfolders/<group-folder-id>` so `getInternalPath()` function returns empty string which resulted into the filename being empty so this PR fixes that as well.


SwikritiT:
> Thanks @SwikritiT! It resolves the issue. Would you like to cover it with a test? I think it is a good idea to have the case covered, but I do not know how much effort it requires. If it is too much then feel free to merge the PR.

yes sure I can cover it with tests

individual-it (approved):
LGTM

But we should actually document how to run the API tests. After this change the group folder app will be needed to run the test.
The documentation should be done in an other PR

SagarGi (approved):
LGTM :+1:

PR #469, opened 2023-08-23
[OP#49775] Drop support for Nextcloud 24

SwikritiT:
Related work package [OP#49775] https://community.openproject.org/projects/nextcloud-integration/work_packages/49775

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,17 +15,8 @@ jobs:
     name: unit tests and linting
     strategy:
       matrix:
-        nextcloudVersion: [ stable24, stable25, stable26, stable27, master ]
-        phpVersion: [ 7.4, 8.0, 8.1 ]
-        exclude:
-          - nextcloudVersion: stable24
-            phpVersion: 8.1
-          - nextcloudVersion: stable26
-            phpVersion: 7.4
-          - nextcloudVersion: stable27
-            phpVersion: 7.4
-          - nextcloudVersion: master
-            phpVersion: 7.4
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -15,11 +15,9 @@ jobs:
     name: unit tests and linting
     strategy:
       matrix:
-        nextcloudVersion: [ stable24, stable25, stable26, stable27, master ]
+        nextcloudVersion: [ stable25, stable26, stable27, master ]
         phpVersion: [ 7.4, 8.0, 8.1 ]
```
individual-it, line 19: I think we can also drop running the tests with PHP 7.4 as its excluded everywhere

tests/lib/Service/DirectDownloadServiceTest.php
```diff
@@ -63,11 +62,8 @@ protected function setUp(): void {
 		/** @var IRootFolder $root */
 		$root = $c->get(IRootFolder::class);
 		$this->userFolder = $root->getUserFolder(self::USER_ID);
-
-		// DirectController constructor changed from NC version 24, see https://github.com/nextcloud/server/pull/32482
-		if (version_compare(OC_Util::getVersionString(), '24') >= 0) {
-			// @phpstan-ignore-next-line
-			$this->directController = new DirectController(
+		// @phpstan-ignore-next-line
```
individual-it, line 65: I think phpstan should be happy without that line now

individual-it (changes requested):

individual-it (approved):

PR #475, opened 2023-08-25
[WP#49787]Remove the function for emptying search input

SagarGi:
### Description:
Forward port of this PR: https://github.com/nextcloud/integration_openproject/pull/474 to `master` branch


individual-it (approved):
Generally its OK to forward port things, but also everything will be forward ported on release

PR #476, opened 2023-08-25
[WP#49786]Remove border and check for consistency in `WorkPackages`

SagarGi:
### Description
The earlier multi select did not had a border line when navigating with keyboard. But the new `NcSelect` has it own border when navigating with keyboard which made the inconsistency in the UI.

### Before With Border

With Keyboard-navigation:

keyboard-action-lightmode

border-fix

### After without border (But with Default NcSelect Border)

keyboard-light-mode

keyboard-darkmode

### Related WorkPaakage
https://community.openproject.org/projects/nextcloud-integration/work_packages/49786/activity?query_id=3787

### DIFF
--- a/src/components/tab/WorkPackage.vue
+++ b/src/components/tab/WorkPackage.vue
@@ -225,7 +225,6 @@ export default {
 .workpackage {
 	width: 100%;
 	padding: 15px 6px 0 10px;
-	border-bottom: 1px solid rgb(237 237 237);
 
 	.row {
 		display: flex;


SagarGi:
Hi @wielinde. I have remove the static border which made inconsistency in hover and also keyboard navigation to WPs. The only thing i see when it is removed is that it does not distinctly separate the workpackages with a boarder like before. Since we are using `NcSelect` it has its own border. And also the other apps in `Nextcloud` using the `NcSelect` and not having the distinct custom static border. 

Its seems like this before with border:
before-without-border

Its seems like this before without border in NcSelect:
without-border-new.


Please have a look into it if we need that static border in new `NcSelect` as well.

SwikritiT (approved):
Tested locally. LGTM

PR #479, opened 2023-08-29
[Backport/468] [OP#49754] fix File(s)info endpoint does not return fileID for group folders

SwikritiT:
Backport- https://github.com/nextcloud/integration_openproject/pull/468
[OP#49754] - https://community.openproject.org/work_packages/49754

For groupfolders if the fileId's are sent as string then the request returned `404` and also for group folders the internal path is `__groupfolders/<group-folder-id>` so `getInternalPath()` function returns empty string which resulted into the filename being empty so this PR fixes that as well.


SwikritiT:
@individual-it @SagarGi this needs your approval

SagarGi (approved):

PR #484, opened 2023-09-05
Use node 18 for release

SwikritiT:
Use node 18 for release workflow as the node version we have been using is outdated


individual-it (approved):
:+1: min node version was changed in #380

PR #490, opened 2023-09-06
Disable running tests on master

SwikritiT:
the ci is failing on master due to some groupfolders issue there is a PR that fixes the issue in groupfolders but if that PR doesn't get merged soon maybe we should disable tests in master so that we won't block aby PR's from getting merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,14 +15,12 @@ jobs:
     name: unit tests and linting
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


SwikritiT:
@individual-it @SagarGi I've made this PR to unblock the CI for now. After this https://github.com/nextcloud/groupfolders/pull/2542 PR is merged maybe we can enable master again.

individual-it (approved):

PR #491, opened 2023-09-07
add phpunit tests for linking the workpackage

SwikritiT:
There were no test coverage for this function so add more test covergae


SagarGi (approved):
LGTM :+1:

PR #492, opened 2023-09-07
[OP#49859]  add iframe for Creating work package from within Nextcloud 

SwikritiT:
Related workpackage: https://community.openproject.org/projects/nextcloud-integration/work_packages/49859

Depends on: https://github.com/opf/openproject/pull/13671


SwikritiT:
we won't be implementing iframe PR for creating the work package form is here https://github.com/nextcloud/integration_openproject/pull/500

PR #496, opened 2023-09-12
[WP#40971]Link the multiple selected files in chunks

SagarGi:
### Description
The `OpenProject` server only supports the multiple selected up to only `20` files at once to link to a workpackage. So this PR makes changes in the request to make it support to link the multiple selected files more than `20` in chunks to link to a single workpackage.

### Part of:
https://community.openproject.org/projects/nextcloud-integration/work_packages/40971/activity

### DIFF
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -36,7 +36,6 @@ import axios from '@nextcloud/axios'
 import { generateUrl } from '@nextcloud/router'
 import NcSelect from '@nextcloud/vue/dist/Components/NcSelect.js'
 import WorkPackage from './WorkPackage.vue'
-import { showError, showSuccess } from '@nextcloud/dialogs'
 import '@nextcloud/dialogs/styles/toast.scss'
 import { workpackageHelper } from '../../utils/workpackageHelper.js'
 import { STATE, WORKPACKAGES_SEARCH_ORIGIN } from '../../utils.js'
@@ -159,38 +158,19 @@ export default {
 				return
 			}
 			// since we can link multiple files now we send file information required in an array (whether it's only one value or multiple)
-			let fileInfoForBody = []
-			let successMessage
 			if (this.searchOrigin === WORKPACKAGES_SEARCH_ORIGIN.PROJECT_TAB) {
-				fileInfoForBody.push(this.fileInfo)
-				successMessage = t('integration_openproject', 'Link to work package created successfully!')
-			} else if (this.searchOrigin === WORKPACKAGES_SEARCH_ORIGIN.LINK_MULTIPLE_FILES_MODAL) {
-				fileInfoForBody = this.fileInfo
-				successMessage = t('integration_openproject', 'Links to work package created successfully for selected files!')
-			}
-
-			const config = {
-				headers: {
-					'Content-Type': 'application/json',
-				},
-			}
-
-			const body = {
-				values: {
-					workpackageId: selectedOption.id,
-					fileinfo: fileInfoForBody,
-				},
-			}
-			const url = generateUrl('/apps/integration_openproject/work-packages')
-			try {
-				await axios.post(url, body, config)
-				this.$emit('saved', selectedOption)
-				showSuccess(successMessage)
+				await workpackageHelper.linkFileToWorkPackageWithSingleRequest([this.fileInfo], selectedOption, t('integration_openproject', 'Link to work package created successfully!'), this)
 				this.resetState()
-			} catch (e) {
-				showError(
-					t('integration_openproject', 'Failed to link file to work package')
-				)
+			} else if (this.searchOrigin === WORKPACKAGES_SEARCH_ORIGIN.LINK_MULTIPLE_FILES_MODAL) {
+				if (this.fileInfo.length <= 20) {
+					await workpackageHelper.linkFileToWorkPackageWithSingleRequest(this.fileInfo, selectedOption, t('integration_openproject', 'Links to work package created successfully for selected files!'), this)
+					this.$emit('close', selectedOption)
+					this.resetState()
+				} else {
+					const chunkedFilesInformations = workpackageHelper.chunkMultipleSelectedFilesInformation(this.fileInfo)
+					await workpackageHelper.linkMultipleFilesToWorkPackageWithChunking(chunkedFilesInformations, selectedOption, false, this)
+					this.resetState()
+				}
 			}
 		},
 		async makeSearchRequest(search, fileId) {
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -39,6 +39,7 @@ import WorkPackage from './WorkPackage.vue'
 import '@nextcloud/dialogs/styles/toast.scss'
 import { workpackageHelper } from '../../utils/workpackageHelper.js'
 import { STATE, WORKPACKAGES_SEARCH_ORIGIN } from '../../utils.js'
+import { translate as t } from '@nextcloud/l10n'
 
 const SEARCH_CHAR_LIMIT = 1
 const DEBOUNCE_THRESHOLD = 500
@@ -167,6 +168,7 @@ export default {
 					this.$emit('close', selectedOption)
 					this.resetState()
 				} else {
+					// the selected files will be linked in chunks
 					const chunkedFilesInformations = workpackageHelper.chunkMultipleSelectedFilesInformation(this.fileInfo)
 					await workpackageHelper.linkMultipleFilesToWorkPackageWithChunking(chunkedFilesInformations, selectedOption, false, this)
 					this.resetState()
--- a/src/components/tab/SearchInput.vue
+++ b/src/components/tab/SearchInput.vue
@@ -169,8 +169,7 @@ export default {
 					this.resetState()
 				} else {
 					// the selected files will be linked in chunks
-					const chunkedFilesInformations = workpackageHelper.chunkMultipleSelectedFilesInformation(this.fileInfo)
-					await workpackageHelper.linkMultipleFilesToWorkPackageWithChunking(chunkedFilesInformations, selectedOption, false, this)
+					await workpackageHelper.linkMultipleFilesToWorkPackageWithChunking(this.fileInfo, selectedOption, false, this)
 					this.resetState()
 				}
 			}
--- a/src/utils/workpackageHelper.js
+++ b/src/utils/workpackageHelper.js
@@ -98,7 +98,7 @@ export const workpackageHelper = {
 		}
 		return chunkedResult
 	},
-	getTotalNoOfFilesFromChunks(chunkedFilesInformations) {
+	getTotalNoOfFilesInAlreadyChunkedFilesInformation(chunkedFilesInformations) {
 		const totalFiles = []
 		for (let i = 0; i < chunkedFilesInformations.length; i++) {
 			totalFiles.push(...chunkedFilesInformations[i])
@@ -111,9 +111,9 @@ export const workpackageHelper = {
 	*/
 	async linkMultipleFilesToWorkPackageWithChunking(chunkedFilesInformations, selectedWorkpackage, isRemaining, component) {
 		const chunkingInformation = {
-			totalNoOfFilesSelected: this.getTotalNoOfFilesFromChunks(chunkedFilesInformations),
+			totalNoOfFilesSelected: this.getTotalNoOfFilesInAlreadyChunkedFilesInformation(chunkedFilesInformations),
 			totalFilesAlreadyLinked: 0,
-			totalFilesNotLinked: this.getTotalNoOfFilesFromChunks(chunkedFilesInformations),
+			totalFilesNotLinked: this.getTotalNoOfFilesInAlreadyChunkedFilesInformation(chunkedFilesInformations),
 			isChunkingError: false,
 			remainingFileInformations: [],
 			selectedWorkPackage: selectedWorkpackage,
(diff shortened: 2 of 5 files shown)


SagarGi:
@wielinde @SwikritiT @individual-it I have made some UI changes for the progress of linking. Please share your opinion towards it if something needs to be changed.

## In Prgoress
chunkProgress

## Link Remaining

errorLinking

src/utils/workpackageHelper.js
```diff
@@ -89,4 +90,96 @@ export const workpackageHelper = {
...
+			chunkedResult.push(fileInformation.slice(i, i + 20))
+		}
+		return chunkedResult
+	},
+	getTotalNoOfFilesInAlreadyChunkedFilesInformation(chunkedFilesInformations) {
+		const totalFiles = []
+		for (let i = 0; i < chunkedFilesInformations.length; i++) {
+			totalFiles.push(...chunkedFilesInformations[i])
+		}
+		return totalFiles.length
```
individual-it, line 106: will that not always be the same as `chunkedFilesInformations.length` ?

src/views/LinkMultipleFilesModal.vue
```diff
@@ -68,12 +123,52 @@ export default {
...
+			const selectedWorkPackage = this.chunkingInformation.selectedWorkPackage
+			const chunkedFilesInformations = workpackageHelper.chunkMultipleSelectedFilesInformation(remainingFilesToChunk)
+			await workpackageHelper.linkMultipleFilesToWorkPackageWithChunking(chunkedFilesInformations, selectedWorkPackage, true, this)
+			if (this.getTotalNoOfFilesAlreadyLinkedInChunking !== remainingFilesToChunk.length) {
+				showError(
+					t('integration_openproject', 'Failed to link selected files to work package')
+				)
+			}
+		},
+		getChunkedInformations(data) {
```
individual-it, line 169: isn't that rather a setter?

individual-it (changes requested):

src/utils/workpackageHelper.js
SagarGi: `chunkedFilesInformations` contains array of array as `[[20 filesInfo], [20 filesInfo], [15 filesInfo]]`. This function get the count of those files.

src/components/tab/SearchInput.vue
```diff
@@ -159,38 +159,20 @@ export default {
...
-				)
+			} else if (this.searchOrigin === WORKPACKAGES_SEARCH_ORIGIN.LINK_MULTIPLE_FILES_MODAL) {
+				if (this.fileInfo.length <= 20) {
+					await workpackageHelper.linkFileToWorkPackageWithSingleRequest(this.fileInfo, selectedOption, t('integration_openproject', 'Links to work package created successfully for selected files!'), this)
+					this.$emit('close', selectedOption)
+					this.resetState()
+				} else {
+					// the selected files will be linked in chunks
+					const chunkedFilesInformations = workpackageHelper.chunkMultipleSelectedFilesInformation(this.fileInfo)
+					await workpackageHelper.linkMultipleFilesToWorkPackageWithChunking(chunkedFilesInformations, selectedOption, false, this)
```
individual-it, line 173: these both functions always go together and actually depend on each other, so what about calling `chunkMultipleSelectedFilesInformation` from within `linkMultipleFilesToWorkPackageWithChunking`?

src/utils/workpackageHelper.js
individual-it: OK, then I would suggest not to create a new array just for counting, but count inside of the loop

src/utils/workpackageHelper.js
```diff
@@ -89,4 +90,96 @@ export const workpackageHelper = {
...
+
+	/*
+		Here the chunkedFilesInformation is the array that contains the array of chunked files information in 20
+	*/
+	async linkMultipleFilesToWorkPackageWithChunking(chunkedFilesInformations, selectedWorkpackage, isRemaining, component) {
+		const chunkingInformation = {
+			totalNoOfFilesSelected: this.getTotalNoOfFilesInAlreadyChunkedFilesInformation(chunkedFilesInformations),
+			totalFilesAlreadyLinked: 0,
+			totalFilesNotLinked: this.getTotalNoOfFilesInAlreadyChunkedFilesInformation(chunkedFilesInformations),
+			isChunkingError: false,
```
individual-it, line 117: is this an error about chunking or when the transfer did not happen?
```suggestion
			error: false,
```

src/views/LinkMultipleFilesModal.vue
```diff
@@ -2,9 +2,48 @@
...
+					<div v-if="getChunkStateInChunking" class="link-progress-information-failed">
+						<div class="link-progress-information-failed--details">
+							<div class="link-progress-information-failed--details--info">
+								<AlertCircleOutline fill-color="#e9322d" :size="70" />
+							</div>
+							<div class="link-progress-information-failed--details--info">
+								<p>{{ t('integration_openproject', `Files selected: ${getTotalNoOfFilesSelectedInChunking}`) }}</p>
+								<p>{{ t('integration_openproject', `Files successfully linked: ${getTotalNoOfFilesAlreadyLinkedInChunking}`) }}</p>
+								<p class="link-progress-information-failed--details--info--error">
+									{{ t('integration_openproject', `Files failed to linked: ${getTotalNoOfFilesNotLinkedInChunking}`) }}
```
individual-it, line 19:
```suggestion
									{{ t('integration_openproject', `Files failed to be linked: ${getTotalNoOfFilesNotLinkedInChunking}`) }}
```

src/views/LinkMultipleFilesModal.vue
```diff
@@ -68,12 +123,52 @@ export default {
...
+			const progressPercentage = parseInt((this.getTotalNoOfFilesAlreadyLinkedInChunking / this.getTotalNoOfFilesSelectedInChunking) * 100)
+			if (progressPercentage === 100) {
+				this.closeRequestModal()
+				showSuccess(
+					t('integration_openproject', 'Links to work package created successfully for selected files!')
+				)
+			}
+			return progressPercentage
+		},
+		getChunkStateInChunking() {
```
individual-it, line 145:
```suggestion
		getError() {
```

src/views/LinkMultipleFilesModal.vue
```diff
@@ -68,12 +123,52 @@ export default {
...
+			return this.chunkingInformation?.isChunkingError
+		},
+		canCloseModal() {
```
individual-it, line 148:
```suggestion
		isError() {
```

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -701,120 +702,331 @@ describe('SearchInput.vue', () => {
...
-								workpackageId: 999,
-								fileinfo: multipleFileInfos,
-							},
+				})
+
+				describe('more than 20 with chunk', () => {
+					/*
+						For the test of linking multiple files more than 20
+						This test scenario create a file information of 55 which is used through the whole test for the link with chunking
+						It means the file will get hchunked as [20, 20, 25]
```
individual-it, line 828:
```suggestion
						It means the file will get chunked as [20, 20, 15]
```

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -701,120 +702,331 @@ describe('SearchInput.vue', () => {
...
-								status: 200,
-								data: workPackageSearchReqResponse,
-							}))
-							.mockImplementation(() => Promise.resolve(
-								{ status: 200, data: [] })
-							)
-						await wrapper.setProps({
-							// here already linked package is empty when the file selected files is more than 1
-							linkedWorkPackages: [],
+						it('should emit event "get-chunked-informations" for 3 times', async () => {
```
individual-it, line 889: isn't that pretty much the same as the test above?

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -701,120 +702,331 @@ describe('SearchInput.vue', () => {
...
+							expect(spyOnEmit).toHaveBeenCalledTimes(3)
 						})
 
-						const inputField = wrapper.find(inputSelector)
-						await inputField.setValue('anything longer than 3 char')
-						for (let i = 0; i < 8; i++) {
-							await localVue.nextTick()
-						}
-						expect(wrapper.vm.searchResults).toMatchObject(
+						it('should emit event "get-chunked-informations" with data', async () => {
```
individual-it, line 904: could this be integrated into above test?

tests/jest/views/LinkMultipleFilesModal.spec.js
```diff
@@ -405,12 +410,122 @@ describe('LinkMultipleFilesModal.vue', () => {
...
+		})
+
+		describe('close', () => {
 			it('should closed the modal', async () => {
-				await wrapper.vm.onSaved()
+				await wrapper.vm.closeRequestModal()
 				expect(wrapper.find(ncModalStubSelector).exists()).toBeFalsy()
 			})
 
+			it('should make chunkng information empty', async () => {
```
individual-it, line 516:
```suggestion
			it('should clean chunking information', async () => {
```

tests/jest/views/LinkMultipleFilesModal.spec.js
```diff
@@ -405,12 +410,122 @@ describe('LinkMultipleFilesModal.vue', () => {
...
+				await wrapper.vm.closeRequestModal()
+				expect(wrapper.vm.chunkingInformation).toBe(null)
+			})
+
+			it('should make chunkng information empty when not empty', async () => {
```
individual-it, line 521: can be integrated into above one with `it.each`

individual-it (changes requested):

src/utils/workpackageHelper.js
SagarGi: yes it is basically error key when there is some error while linking with chunk. But may be `error` is enough.

tests/jest/components/tab/SearchInput.spec.js
SagarGi: yes but this checks for the emit event to the parent component

tests/jest/components/tab/SearchInput.spec.js
SagarGi: it is the case of successful link with chunk. I have adjusted the test case description instead.

individual-it (approved):

src/utils/workpackageHelper.js
```diff
@@ -89,4 +90,101 @@ export const workpackageHelper = {
 			fileId: workPackage.fileId,
 		}
 	},
+	chunkMultipleSelectedFilesInformation(fileInformation) {
+		// this function chunks 20 files information in an array and returns array of it
```
SwikritiT, line 94: Can we also add a little bit of context on why exactly  `20` so that we are clear in future as well?
```suggestion
		// this function chunks 20 files of information in an array and returns an array of it, `20` because Openproject api only allows linking of 20 files per request
```

src/utils/workpackageHelper.js
```diff
@@ -89,4 +90,101 @@ export const workpackageHelper = {
...
+		let totalNoFiles = 0
+		for (let i = 0; i < chunkedFilesInformations.length; i++) {
+			totalNoFiles = totalNoFiles + chunkedFilesInformations[i].length
+		}
+		return totalNoFiles
+	},
+
+	/*
+		Here the chunkedFilesInformation is the array that contains the array of chunked files information in 20
+	*/
```
SwikritiT, line 111:
```suggestion
		// Here the chunkedFilesInformation is the array that contains the array of chunked files information in 20
```

src/views/LinkMultipleFilesModal.vue
```diff
@@ -2,9 +2,48 @@
...
 			v-if="show"
+			:can-close="isError"
 			@close="closeRequestModal">
 			<div class="multiple-link-modal-content">
 				<LoadingIcon v-if="isLoading" class="loading-spinner" :size="60" />
+				<div v-else-if="chunkingInformation !== null" class="link-progress-information-wrapper">
+					<div v-if="getError" class="link-progress-information-failed">
+						<div class="link-progress-information-failed--details">
+							<div class="link-progress-information-failed--details--info">
+								<AlertCircleOutline fill-color="#e9322d" :size="70" />
```
SwikritiT, line 13: Can we use the variable of the colours provided by the nextcloud here as those colours are handeled by NC itself in different themes?
```suggestion
								<AlertCircleOutline fill-color="--var(red-color-provided-by-nextcloud)" :size="70" />
```

SwikritiT (changes requested):

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -701,120 +701,318 @@ describe('SearchInput.vue', () => {
...
+							const axiosSpy = jest.spyOn(axios, 'get')
+								.mockImplementationOnce(() => Promise.resolve({
+									status: 200,
+									data: workPackageSearchReqResponse,
+								}))
+								.mockImplementation(() => Promise.resolve(
+									{ status: 200, data: [] })
+								)
+							await wrapper.setProps({
+								// here already linked package is empty when the file selected files is more than 1
```
SwikritiT, line 772:
```suggestion
								// here already linked work package package is empty when the selected files is more than 1
```

tests/jest/components/tab/SearchInput.spec.js
```diff
@@ -701,120 +701,318 @@ describe('SearchInput.vue', () => {
...
+							axiosSpy.mockRestore()
+						})
+					})
+				})
+
+				describe('more than 20 with chunk', () => {
+					/*
+						For the test of linking multiple files more than 20
+						This test scenario create a file information of 55 which is used through the whole test for the link with chunking
+						It means the file will get chunked as [20, 20, 15]
```
SwikritiT, line 827:
```suggestion
						For the test of linking multiple files more than 20.
						This test scenario creates a file information of 55 which is used through the whole test for the link with chunking.
						It means the file will get chunked as [20, 20, 15].
```

SwikritiT:
Note that in nextcloud master i.e NC-28 seems like the design has changed and the multi-linking option is nowhere to be seen.

Screenshot from 2023-10-18 14-23-23




 If we want this feature for 2.5 and if https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule this is correct nextcloud 28 is set to be released around December so we need to make it compatible with nextcloud 28 as well.

PR #497, opened 2023-09-21
[WP#50137]Remove check for class exists for `groupfolders`

SagarGi:
### Description
The check for class `class_exists('\OCA\GroupFolders\Folder\FolderManager')` in function `isGroupfoldersAppEnabled` seems to be useless since it is `true` when app is enabled/installed and `false` when app is disabled/removed. So Error may arise when making an extra check by it since all the thing can be checked by `isEnabledForUser` this function check enabled/installed or disable for the `groupfolder` app

### Related WP
https://community.openproject.org/projects/nextcloud-integration/work_packages/50137/activity?query_id=4615

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -1090,7 +1090,6 @@ class OpenProjectAPIService {
 	public function isGroupfoldersAppEnabled(): bool {
 		$user = $this->userManager->get(Application::OPEN_PROJECT_ENTITIES_NAME);
 		return (
-			class_exists('\OCA\GroupFolders\Folder\FolderManager') &&
 			$this->appManager->isEnabledForUser(
 			'groupfolders',
 			$user


SagarGi:
@SwikritiT @individual-it The function `isEnabledForUser` checks whether the app is enabled or disabled. May be only this functiion can be enough to check.

individual-it (commented):
LGTM, please check with @wielinde if this fixes the issue

wielinde:
@SagarGi @individual-it Unfortunately I don't have a local replication of that bug. Clients that had installed the app through the app store observed the bug. I am afraid we cannot checkout the branch of this PR on the client's machine. However, if you guys verified that the code still works (enabling/disabling/installing/uinstalling the "Group folders" app) then we might consider simply releasing a patch release. Or do you have a better idea?

SagarGi:
> @SagarGi @individual-it Unfortunately I don't have a local replication of that bug. Clients that had installed the app through the app store observed the bug. I am afraid we cannot checkout the branch of this PR on the client's machine. However, if you guys verified that the code still works (enabling/disabling/installing/uinstalling the "Group folders" app) then we might consider simply releasing a patch release. Or do you have a better idea?

if its specifically encountered installing from the app store then i can try to reproduce it locally first.

PR #499, opened 2023-09-22
[OP#50153 ] Provide logger as parametr while creatimg the Data object in FilesController

SwikritiT:
related work package [OP#50153] : https://community.openproject.org/projects/nextcloud-integration/work_packages/details/50153

In nextcloud, master new parameter i.e LoggerInterface is needed while creating the object of  `Data` class so this PR adds that. This solves the failure that we have been seeing in API tests. 

> Note: we also have ILogger which is deprecated but some of the nextcloud functions still expect `ILogger` so we cannot exchange `ILooger` with `LoggerInterface` for now and both are needed as per different NC versions that we are supporting.


lib/Controller/FilesController.php
```diff
@@ -227,7 +254,9 @@ private function compileFileInfo($fileId) {
 				);
 				$internalPath = $file->getName();
 			}
-			$modifier = $this->getLastModifier($owner->getUID(), $file->getId());
+			$a = $owner->getUID();
+			$b = $file->getId();
```
SwikritiT, line 258: there are some debugging leftovers, I'll remove them once this CI is green.

individual-it (approved):

PR #502, opened 2023-10-12
fix: Use DI for getting 3rdparty app classes

juliushaertl:
Fix incompatible constructor arguments for the activity classes by using the dependency injection container get method to automatically construct it.

Relevant upstream change https://github.com/nextcloud/activity/pull/1309 but the adjustment here makes it less error prone for further changes.

# Related WorkPackage:
This PR fixes: https://community.openproject.org/projects/nextcloud-integration/work_packages/50507/activity

### DIFF
--- a/lib/Controller/FilesController.php
+++ b/lib/Controller/FilesController.php
@@ -23,9 +23,7 @@ use OCP\Files\Config\ICachedMountFileInfo;
 use OCP\Files\Config\IMountProviderCollection;
 use OCP\Files\FileInfo;
 use OCP\Files\IRootFolder;
-use OCP\IConfig;
 use OCP\IDBConnection;
-use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IRequest;
 use OCP\AppFramework\OCSController;
@@ -34,7 +32,7 @@ use OCP\AppFramework\Http\DataResponse;
 use OCP\IUser;
 use OCP\IUserManager;
 use OCP\IUserSession;
-use OCP\RichObjectStrings\IValidator;
+use OCP\Server;
 use Throwable;
 use OCP\Files\DavUtil;
 
@@ -61,24 +59,9 @@ class FilesController extends OCSController {
 	protected $connection;
 
 	/**
-	 * @var IValidator
-	 */
-	private $richObjectValidator;
-
-	/**
 	 * @var ILogger
 	 */
 	private $logger;
-
-	/**
-	 * @var IL10N
-	 */
-	private $l;
-
-	/**
-	 * @var IConfig
-	 */
-	private $config;
 
 	/**
 	 * @var IUserManager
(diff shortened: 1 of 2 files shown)


juliushaertl:
Not sure about the CI failures, seem to also be the case on the master branch

SwikritiT:
> Not sure about the CI failures, seem to also be the case on the master branch

yes that's related to pslam.

wielinde:
@juliushaertl Thank you for the PR. Could you please base the PR against the branch `release/2.4`? That is the branch for the next patch level release of the plugin.

juliushaertl:
@wielinde Done. Now based on `release/2.4`

juliushaertl:
I also amended a bit more cleanup of properties that are no longer needed.

SwikritiT:
I've pushed a commit to fix the unit tests https://github.com/nextcloud/integration_openproject/pull/502/commits/6b515275a8128071d3d906446a4f77261feff1c4
The CI is green when it is disabled in master https://github.com/nextcloud/integration_openproject/actions/runs/6503810966

SwikritiT (approved):

wielinde:
@SwikritiT did anybody also test this PR on the older, supported Nextcloud versions? If yes, please release it as a patch level release

SwikritiT:
> @SwikritiT did anybody also test this PR on the older, supported Nextcloud versions? If yes, please release it as a patch level release

We run the tests on all the supported versions and the API tests are passing in all of them. But we can do a quick manual test on all the versions as well

wielinde:
> > > @SwikritiT did anybody also test this PR on the older, supported Nextcloud versions? If yes, please release it as a patch level release
> > 
> > 
> > We run the tests on all the supported versions and the API tests are passing in all of them. But we can do a quick manual test on all the versions as well
> 
> Did the manual test and everything works on nextcloud side, in all the supported versions. I'll merge this PR and release the patch level

Cool. Thank you for the QA.

One question out of curiosity: Did our CI fail when Nextcloud released 27.1.2? I mean, did it fail before this PR? I guess that the CI only runs on changes in the app. So if Nextcloud has a new stable release, that makes our CI fail, then we will not notice until our next commit to the app, which might be too late. Maybe we need to pick up on changes on Nextcloud stable branches, too?

SwikritiT:
> > > > @SwikritiT did anybody also test this PR on the older, supported Nextcloud versions? If yes, please release it as a patch level release
> > > 
> > > 
> > > We run the tests on all the supported versions and the API tests are passing in all of them. But we can do a quick manual test on all the versions as well
> > 
> > 
> > Did the manual test and everything works on nextcloud side, in all the supported versions. I'll merge this PR and release the patch level
> 
> Cool. Thank you for the QA.
> 
> One question out of curiosity: Did our CI fail when Nextcloud released 27.1.2? I mean, did it fail before this PR? I guess that the CI only runs on changes in the app. So if Nextcloud has a new stable release, that makes our CI fail, then we will not notice until our next commit to the app, which might be too late. Maybe we need to pick up on changes on Nextcloud stable branches, too?

Yes, we do pull the changes from the stable branches to run the CI so I would expect to contain the changes for the release(I'll double-check to be sure). But like you said currently the build only runs when there are new changes through PR or push. Sometimes the changes in the PR might cause the CI to fail so some failures like this case might not be noticed until too late. Maybe we need to run nightly builds in release and maybe the master branch which will be more reliable in catching these kinds of errors and the CI.

PR #507, opened 2023-10-17
[WP#50135]Fix the delete files in shared link

SagarGi:
### Description
Fixes : https://community.openproject.org/projects/openproject/work_packages/50135/activity

### DIFF
--- a/lib/Listener/BeforeNodeInsideOpenProjectGroupfilderChangedListener.php
+++ b/lib/Listener/BeforeNodeInsideOpenProjectGroupfilderChangedListener.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace OCA\OpenProject\Listener;
 
+use OC_User;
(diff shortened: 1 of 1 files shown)


lib/Listener/BeforeNodeInsideOpenProjectGroupfilderChangedListener.php
```diff
@@ -50,6 +51,9 @@ public function handle(Event $event): void {
 		} else {
 			return;
 		}
+		if (OC_User::isIncognitoMode()) {
+			return;
```
SwikritiT, line 55: Can we also add a comment here on why this is needed?

SwikritiT (changes requested):
Just figured out that with these changes if the OpenProject group folder is shared through public link then the nodes inside it can be deleted.

SwikritiT (approved):

individual-it (approved):

PR #508, opened 2023-10-17
[#50431] Project folders setup: add special characters to generated passwords

SwikritiT:
Fixes: https://community.openproject.org/work_packages/50431
As the password that we generated for project folders didn't contain the special characters in case the nextcloud instance has a password policy setup enforcing special characters, an error is shown while setting up the project folders. This PR adds special characters in the password to prevent that from happening


individual-it (approved):
I'm surprised that an app password also has to meet those requirements.
Please check if there are more requirements that the password policy can enforce and we would not meet.

SagarGi (approved):
LGTM :+1:

SwikritiT:
> I'm surprised that an app password also has to meet those requirements. Please check if there are more requirements that the password policy can enforce and we would not meet.

These is the password policy options in all the supported versions of next cloud, I think or app fulfills it
Screenshot from 2023-10-18 11-38-18

PR #514, opened 2023-10-30
[OP#50617] Use dependecy Injection to get classes from other apps

SwikritiT:
We have some functions that require the classes of other nextcloud apps like groupfolders and as we are supporting 3 major versions of the nextcloud server, the parameters to the class constructors may change according to the server's version. Currently, we are determining the version of the nextcloud and providing parameters based on that but in this PR we use dependency injection like in PR https://github.com/nextcloud/integration_openproject/pull/502

Related work package: https://community.openproject.org/projects/nextcloud-integration/work_packages/50617


individual-it (approved):
This makes the code much nicer :+1:

PR #515, opened 2023-11-03
[WP#50814]Look up for to migrate to new files action register

SagarGi:
### Description
The files action for the integration app for the new version `NC-28` or greater is registered in a style different from older versions. So a separate script is needed to be loaded for it.

### Relates WP:
https://community.openproject.org/projects/nextcloud-integration/work_packages/50814
https://community.openproject.org/projects/nextcloud-integration/work_packages/50608


julien-nc:
I pushed a commit that changes the event listening to load the script (that was not loaded before).
I changed the webpack config to be able to load svg files without `?raw` and without the eslint warning.
It works now.
There is still one bug of the file action system (in the core, nothing on you) that fires an error even if the execBatch function is successful. This has to be fixed on the NC side.

julien-nc:
Just in case: To set a different action name on the single and multiple menu entry: https://github.com/nextcloud/integration_mattermost/blob/main/src/filesplugin.js#L56-L60

SagarGi:
> Just in case: To set a different action name on the single and multiple menu entry: https://github.com/nextcloud/integration_mattermost/blob/main/src/filesplugin.js#L56-L60

We need to find the condition for what the user has selected the file through may be. (context menu or the batch action)

individual-it (approved):

SwikritiT (changes requested):
Needs some more adjustments. Requesting changes to prevent accidental merging

SwikritiT:
The UI looks as expected and linking is also working

individual-it (approved):

PR #519, opened 2023-11-20
Improve debuggability and minor fixes

rossnet:
I use this script to bootstrap the OpenProject - Nextcloud connection within a K8s deployment containing both instances. The applied changes help a lot when i comes to debugging the actual setup sequence.


SwikritiT:
@SagarGi @individual-it I think we can review and merge this if everything is working. Since this is from fork CI won't pass and the changes shouldn't affect CI either.

individual-it (approved):

SagarGi (approved):
Checked it locally working fine. LGTM :+1:

SagarGi (changes requested):
Fails when running the script with setting up the group folders through it.

PR #521, opened 2023-11-24
Make CI run with stable28

SwikritiT:
Nextcloud checked out a stable-28 branch https://github.com/nextcloud/server/tree/stable28 so now the master is version-29 , since apps like `groupfolders` aren't yet compatible with nextcloud version-29 we are getting CI failures like

```console
Run git clone --depth 1 https://github.com/nextcloud/groupfolders.git -b master server/apps/groupfolders
Cloning into 'server/apps/groupfolders'...
App "Group folders" cannot be installed because it is not compatible with this version of the server.
Error: Process completed with exit code 1.
```

This will unblock the CI.

## Related Workpackage
OP#51218 - https://community.openproject.org/projects/nextcloud-integration/work_packages/51218

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,7 @@ jobs:
           sudo apt install make openssl -y
           echo "###### installing nextcloud"
           mkdir ~/html
-          git clone https://github.com/nextcloud/server.git --recursive --depth 1 -b master ~/html/nextcloud
+          git clone https://github.com/nextcloud/server.git --recursive --depth 1 -b stable28 ~/html/nextcloud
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           sudo apt install make openssl -y
           echo "###### installing nextcloud"
           mkdir ~/html
-          git clone https://github.com/nextcloud/server.git --recursive --depth 1 -b master ~/html/nextcloud
+          git clone https://github.com/nextcloud/server.git --recursive --depth 1 -b stable28 ~/html/nextcloud
```
SwikritiT, line 59: I've also updated release workflow to use `stable28` , I can revert this if we will support 29 for our upcoming release of this app

SagarGi (approved):
LGTM :+1:

individual-it (approved):
good!
@SagarGi please create an follow-up WP, so that we don't forget to go back to `master`

PR #524, opened 2023-11-29
[WP#50542]Added job for sending nightly report to element with script

SagarGi:


.github/scripts/notify-to-element.sh
```diff
@@ -0,0 +1,56 @@
...
+
+log_info() {
+  echo -e "\e[37m$1\e[0m"
+}
+
+log_success() {
+  echo -e "\e[32m$1\e[0m"
+}
+
+log_info "Fetching the all workflow jobs....."
```
individual-it, line 16:
```suggestion
log_info "Fetching all workflow jobs....."
```

individual-it (approved):

PR #527, opened 2023-11-30
[OP#]fix: Virus infected File is not removed by background job when integration app is enabled

SwikritiT:
Related Issues:
- https://community.openproject.org/work_packages/51321
- https://github.com/nextcloud/files_antivirus/issues/297


lib/Listener/BeforeNodeInsideOpenProjectGroupfilderChangedListener.php
```diff
@@ -51,11 +51,13 @@ public function handle(Event $event): void {
 		} else {
 			return;
 		}
+		$currentUser = $this->userSession->getUser();
 		// we do not listen event where user is not logged or there is no user session (e.g. public link )
-		if (OC_User::isIncognitoMode()) {
+		// or if it's some background job in which case user will be null
+		if (OC_User::isIncognitoMode() || $currentUser === null) {
 			return;
 		}
```
SwikritiT, line 59: I've checked if the user is null and if the user is null then it's a "background job " . I couldn't find a way to determine if we can know an event is triggered by a background job. @julien-nc would you have any suggestion on if I can determine that a task is done by background job?

individual-it (approved):
That looks good to me.
@wielinde can you ask the people who reported the issue in the first place to test this fix?

wielinde:
Hey @Rello, could you please verify that this PR solves the problem? Thank you!

SwikritiT:
Hi, @wielinde @Rello any updates on if we can move ahead with the changes of this PR?

Rello:
will test today...

Rello:
@SwikritiT just tested and confirmed working 👍🏻

wielinde:
Thank you @Rello! Much apreciated.

Now, that this is fixed, please allow me this little comment: Nextcloud could really benefit from a concept that some objects, such as files, folders, groups, users, group folders are _owned_ by an application and can get protected from getting deleted or modified by somebody else. Then, we would not need to hook into every deletion every file, folder, group and user in order to prevent them from being deleted by some uninformed admin. And still a app could use those objects for creating cool work flow optimization such as our project folder concept.

Rello:
> Thank you @Rello! Much apreciated.
> 
> Now, that this is fixed, please allow me this little comment: Nextcloud could really benefit from a concept that some objects, such as files, folders, groups, users, group folders are _owned_ by an application and can get protected from getting deleted or modified by somebody else. Then, we would not need to hook into every deletion every file, folder, group and user in order to prevent them from being deleted by some uninformed admin. And still a app could use those objects for creating cool work flow optimization such as our project folder concept.

Hallo @wielinde 
thank you for your suggestion. Problem is here, that this would result in objects that are not owned by a real user and this might create auditing challenges. With everything tied to a user, its clear who maintained or changed what content of the system. When switching to alternatives with "app user" or "service user", the whole setup will get very complex.

PR #535, opened 2023-12-06
Make the min password length dynamic instead of hardcoded one

SagarGi:
Forward port of PR: https://github.com/nextcloud/integration_openproject/pull/530

### DIFF
--- a/lib/Service/OpenProjectAPIService.php
+++ b/lib/Service/OpenProjectAPIService.php
@@ -1123,7 +1123,7 @@ class OpenProjectAPIService {
 	 * @return int
 	 */
 	public function getPasswordLength(): int {
-		$passLengthSet = (int) $this->config->getAppValue('password_policy', 'minLength', '');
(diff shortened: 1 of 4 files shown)


lib/Service/OpenProjectAPIService.php
```diff
@@ -1119,13 +1119,21 @@ private function isGroupfolderAppCorrectlySetup():bool {
 		return false;
 	}
 
+	/**
+	 * @return int
+	 */
+	public function getPasswordLength(): int {
+		$passLengthSet = (int) $this->config->getAppValue('password_policy', 'minLength', '');
```
SwikritiT, line 1126: should we add 72 here so that the default will be 72 if the value is not set at all.
```suggestion
		$passLengthSet = (int) $this->config->getAppValue('password_policy', 'minLength', 72);
```

SagarGi: i have added default to 0 and any ways it will be if less than 72 the function will return lenght 72

SwikritiT (approved):

PR #554, opened 2024-01-09
Remove the trashed state from fileId endpoint

SwikritiT:
Related workpackage: https://community.openproject.org/projects/nextcloud-integration/work_packages/52045

### DIFF
--- a/tests/acceptance/features/api/getFileinfoByFileIDAPI.feature
+++ b/tests/acceptance/features/api/getFileinfoByFileIDAPI.feature
@@ -22,7 +22,6 @@ Feature: retrieve file information of a single file, using the file ID
       "owner_name",
       "modifier_id",
       "modifier_name",
-      "trashed",
       "dav_permissions",
       "path"
       ],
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/api/getFileinfoByFileIDAPI.feature
```diff
@@ -214,7 +190,6 @@ Feature: retrieve file information of a single file, using the file ID
       "owner_name",
       "modifier_id",
       "modifier_name",
-      "trashed",
```
individual-it, line 217: let's put that in 
```
"not": {
      "required": [
```
to make sure the parameter is really gone.
Similar for the other tests

tests/acceptance/features/api/getFileinfoByFileIDAPI.feature
```diff
@@ -164,26 +153,13 @@ Feature: retrieve file information of a single file, using the file ID
...
```
individual-it, line 167: let's leave that line here, as its in not-required, that will make sure its really deleted

individual-it (changes requested):
looks very good, just tiny suggestions for the acceptance tests

Kharonus (approved):
LGTM.

From what I can tell, the functionality change is exactly as we discussed, hence accepted from our side.

individual-it (approved):

PR #555, opened 2024-01-15
draft check for change log details

SagarGi:
### Description
The App details in the `nextcloud` app store seems to be according to the change log file. This PR adds just a change log for version `2.5.1`. It will be followed by a Publish nightly release to appstore to check if it works.

### Related WP
https://community.openproject.org/projects/nextcloud-integration/work_packages/52112


SagarGi:
Accidently checked out from release/2.5 instead of release/2.6 so closing this PR for now.

PR #588, opened 2024-03-11
Refactor the runningFullReset varaible

SagarGi:
## Description
The full reset of the integration means resetting all the values that has been set already default null. Some how the `runningFullReset` was resulting `false` even when we were resetting the whole integration. It was because of `$oldClientSecret` `$oldClientId` which made it `false`. These 2 checks has been removed since it seems to be non relevant to resetting the integration.

- Fixes

## Types of changes

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/Controller/ConfigController.php
+++ b/lib/Controller/ConfigController.php
@@ -263,13 +263,13 @@ class ConfigController extends Controller {
 
 		$runningFullReset = (
 
-			$oldClientSecret &&
-
-			$oldClientId &&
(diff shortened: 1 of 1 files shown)


individual-it (approved):

individual-it (commented):
some tests would be good if they make sense

SagarGi:
> some tests would be good if they make sense

We have tests for resetting. But the i think we cannot test for that single variable.

PR #552, opened 2023-12-22
Added TOS app enabled and set TOS for OP user

SagarGi:
### Description
This PR adds:
1. Sign TOS for user OP (created already) when app `terms_of_service` is enabled.
2. Sign TOS for user OP (while creating it) given that `terms_of_service` is already enabled.
3. Sign TOS for user OP when resetting TOS if the user OP is already created
4. Sign TOS for user OP when new TOS is created given that `terms_of_service` and OP is already enabled and created
5. Sign unsigned  TOS for user OP when `integration_openproject` app is disabled and enabled again.

### PR for events made on `terms_of_service` app
https://github.com/nextcloud/terms_of_service/pull/895 in `terms_of_service` app

TODO:
- [x] Add unit tests UI tests
- [x] Add Php unit tests 

### Related Issue:
https://community.openproject.org/projects/nextcloud-integration/work_packages/51804

PR #595, opened 2024-03-14
Remove only from admin setting tests

SagarGi:
## Description
The admin setting test was not fully running because of a test with  `.only` tests which was merged accidentally with it in PR https://github.com/nextcloud/integration_openproject/pull/552/files#diff-437209b827b3b0af71027912f456e4ef6a3b7340eb524e754ac53b12e9b4f72dR1792.

- This PR removes the .only and rerun other tests
- This PR adjust the failed tests due to new changes
- This PR adjust the terms of services sign flag in a `data` rather than directly passing from the `loaded state` into the template.

- Fixes <issue_link>

## Types of changes

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/src/components/AdminSettings.vue
+++ b/src/components/AdminSettings.vue
@@ -425,6 +425,7 @@ export default {
 			// pointer for which form the request is coming
 			isFormStep: null,
 			isDarkTheme: null,
+			isAllTermsOfServiceSignedForUserOpenProject: true,
 		}
 	},
 	computed: {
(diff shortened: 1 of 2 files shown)


individual-it:
I don't quite understand why we need the third point

SagarGi:
> I don't quite understand why we need the third point

the state of openproject user has signed the terms of service was passed directly through the state. We needed to pass it for each state in the tests. Instead a data value was created for it or else we need to pass the  `all_terms_of_services_signed` for each test  we run since previously there was not default value for it.

individual-it (approved):

PR #624, opened 2024-04-10
[WP#48598]Drop support for php7.4 and clean up code

SagarGi:
## Description
This PR clean up code related to `php7.4` support to run tests in CI.

https://community.openproject.org/projects/nextcloud-integration/work_packages/48598/activity

## Types of changes

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/Reference/WorkPackageReferenceProvider.php
+++ b/lib/Reference/WorkPackageReferenceProvider.php
@@ -163,7 +163,7 @@ class WorkPackageReferenceProvider extends ADiscoverableReferenceProvider {
 	 */
 	public function getCacheKey(string $referenceId): ?string {
 		$wpId = $this->getWorkPackageIdFromUrl($referenceId);
-		return $wpId ?? $referenceId;
+		return (string) ($wpId ?? $referenceId);
 	}
 
 	/**
--- a/lib/Reference/WorkPackageReferenceProvider.php
+++ b/lib/Reference/WorkPackageReferenceProvider.php
@@ -163,7 +163,7 @@ class WorkPackageReferenceProvider extends ADiscoverableReferenceProvider {
 	 */
 	public function getCacheKey(string $referenceId): ?string {
 		$wpId = $this->getWorkPackageIdFromUrl($referenceId);
-		return (string) $wpId ?? $referenceId;
+		return $wpId ?? $referenceId;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


lib/Reference/WorkPackageReferenceProvider.php
```diff
@@ -163,7 +163,7 @@ public function getCachePrefix(string $referenceId): string {
 	 */
 	public function getCacheKey(string $referenceId): ?string {
 		$wpId = $this->getWorkPackageIdFromUrl($referenceId);
-		return (string) $wpId ?? $referenceId;
+		return $wpId ?? $referenceId;
```
individual-it, line 166: why doing that?
`getWorkPackageIdFromUrl()` returns an int or null, but `getCacheKey` has to return a string or null. So this change should throw somewhere an error IMO

tests/lib/Reference/WorkPackageReferenceProviderTest.php
```diff
@@ -43,23 +43,23 @@ protected function setUp(): void {
...
 	public function getWorkReferenceProviderMock(
-		array $onlyMethods = ['request'],
-		$configMock = null,
-		$iL10N = null,
-		$iURLGenerator = null,
-		$refrenceMangager = null,
-		$openProjectAPIService = null,
-		$userId = null
+		array                 $onlyMethods = ['request'],
+		MockObject            $configMock = null,
```
individual-it, line 57: I think we should be able to do 
```suggestion
		?MockObject            $configMock = null,
```
and then we can even delete the PHPDoc section and don't have to maintain both

individual-it (changes requested):

lib/Reference/WorkPackageReferenceProvider.php
SagarGi: True. Will change according to it.

tests/lib/Reference/WorkPackageReferenceProviderTest.php
SagarGi: Good to have PHPDoc but instead i cleared up the Type in parameters.

SagarGi: I have created a `workpackage` related to updating the code according to php 8.0. This can be done as house keeping work.

individual-it (approved):

PR #637, opened 2024-05-06
Update script

SagarGi:
## Description
This PR updates the integration script. Previously the integration script used to fails when the integration setup is already completed and re ran. 

- But now there are some checks implemented if it is already completed then it does not give an error but rather output it as a integration setup completed.
- Also an API for admin is added to get if the admin config is completed with project folder setup or without which is used in the script itself.

- https://community.openproject.org/projects/nextcloud-integration/work_packages/54644

## Types of changes

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/integration_setup.sh
+++ b/integration_setup.sh
@@ -217,7 +217,7 @@ if [[ ${response_type} == "Error" ]]; then
     log_info "OPENPROJECT_AUTHENTICATION_GLOBAL__BASIC__AUTH_USER=<basic_auth_api_username>"
     log_info "OPENPROJECT_AUTHENTICATION_GLOBAL__BASIC__AUTH_PASSWORD=<basic_auth_api_password>"
   else
-    log_error "Unhandled error while create file storage '${OPENPROJECT_STORAGE_NAME}'"
+    log_error "Unhandled error while creating the file storage '${OPENPROJECT_STORAGE_NAME}'"
     log_error "OpenProject returned the following error: '${error_message}'"
     log_info "You could try deleting the file storage in OpenProject and run the script again."
   fi
--- a/integration_setup.sh
+++ b/integration_setup.sh
@@ -91,9 +91,9 @@ isNextcloudAdminConfigOk() {
                   -H 'X-Requested-With: XMLHttpRequest')
     admin_config_status=$(echo $admin_config_response | jq -r ".status")
   if [[ ${admin_config_status} == 'true' ]]; then
-      return 0
+      echo 0
   else
-      return 1
+      echo 1
   fi
 }
 
@@ -135,7 +135,6 @@ else
   setup_method=POST
 fi
 
-log_info "Creating file storage name \"${OPENPROJECT_STORAGE_NAME}\" in OpenProject ..."
 # api call to get openproject_client_id and openproject_client_secret
 cat >${INTEGRATION_SETUP_TEMP_DIR}/request_body_1_op_create_storage.json <<EOF
 {
@@ -168,10 +167,33 @@ if [[ ${response_type} == "Error" ]]; then
   if [[ ${error_id} == "urn:openproject-org:api:v3:errors:MultipleErrors" ]]; then
     # if files storage is already created with the provided hostname (nextcloud) and storage name
     # we assume that the integration set up is done in both application
-    error_messages=$(echo $create_storage_response | jq -r "._embedded.errors[]._embedded.details.attribute")
-    if [[ " ${error_messages[*]} " =~ "host" || " ${error_messages[*]} " =~ "name" ]]; then
-        echo "Both exist!!"
+    error_messages=($(echo $create_storage_response | jq -r "._embedded.errors[]._embedded.details.attribute"))
+    count=0
+    for element in "${error_messages[@]}"; do
+      if [[ "$element" == "host" ]]; then
+          (( count +=1 ))
+      elif [[ "$element" == "name" ]]; then
+          (( count +=1 ))
+      fi
+    done
+    if [[ "$count" -ne 2 ]]; then
+    	log_error "Got multiple errors when setting up
+              OP storage: ${OPENPROJECT_STORAGE_NAME}
+              for Nextcloud: '${NEXTCLOUD_HOST}
+              using endpoint: ${OPENPROJECT_BASEURL_FOR_STORAGE}"
+		log_error "Error Message(s): ${error_messages}"
+		log_info "You could try deleting the file storage in OpenProject and run the script again."
+    	exit 1
     fi
+    log_success "File storage with name '${OPENPROJECT_STORAGE_NAME}' with host '${NEXTCLOUD_HOST}' has already been created on 'OpenProject'."
+    status=$(isNextcloudAdminConfigOk)
+    if [[ "$status" -ne 0 ]]; then
+    	log_error "Some admin config for integration is missing on Nextcloud '${NEXTCLOUD_HOST}'."
+    	log_info "You could try deleting the file storage '${OPENPROJECT_STORAGE_NAME}' in OpenProject and run the script again."
+        exit 1
+    fi
+    log_success "Admin config on Nextcloud is already setup."
+    log_success "Setting of the integration was already successful!"
     exit 0
   elif [[ ${error_id} == "urn:openproject-org:api:v3:errors:Unauthenticated" ]]; then
     log_error "Authorization failed. Ensure you have created a valid BASIC AUTH API account, e.g. utilising the following env variables:"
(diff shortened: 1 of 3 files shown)


lib/Service/OauthService.php
```diff
@@ -92,6 +92,7 @@ public function getClientInfo(int $id): ?array {
 				'nextcloud_oauth_client_name' => $client->getName(),
 				'openproject_redirect_uri' => $client->getRedirectUri(),
 				'nextcloud_client_id' => $client->getClientIdentifier(),
+				'nextcloud_client_secret' => $this->crypto->decrypt($client->getSecret()),
```
individual-it, line 95: why do we need this?

integration_setup.sh
```diff
@@ -83,6 +83,27 @@ openproject_host_state_response=$(curl -s -X GET -u${OP_ADMIN_USERNAME}:${OP_ADM
...
 setup_app_password=false
 
+isNextcloudAdminConfigOk() {
+	admin_config_response=$(curl -s -X GET -u${NC_ADMIN_USERNAME}:${NC_ADMIN_PASSWORD} \
+				  ${NC_INTEGRATION_BASE_URL}/check-admin-config \
+				  -H 'accept: application/hal+json' \
+				  -H 'Content-Type: application/json' \
+				  -H 'X-Requested-With: XMLHttpRequest')
+	admin_config_status_without_project_folder=$(echo $admin_config_response | jq -r ".status_without_project_folder_setup_status")
+	admin_config_status_with_project_folder=$(echo $admin_config_response | jq -r ".status_with_project_folder_setup_status")
```
individual-it, line 93: maybe I don't understand the naming but it sounds to me that these values should be opposite of each other, so if `admin_config_status_with_project_folder == true` then `admin_config_status_without_project_folder == false`?

lib/Controller/ConfigController.php
```diff
@@ -557,6 +557,22 @@ public function checkConfig(): DataResponse {
...
+	 */
+	public function checkAdminConfigOk(): DataResponse {
+		$adminConfigStatusWithoutGroupFolderSetupStatus = OpenProjectAPIService::isAdminConfigOk($this->config);
+		$appPasswordSetStatus = $this->openprojectAPIService->hasAppPassword();
+		// admin config completion can be checked in 2 phase
+		// 1. with project folder setup (with app password)
+		// 2. without project folder setup (no app password)
+		return new DataResponse([
+			'status_without_project_folder_setup_status' => $adminConfigStatusWithoutGroupFolderSetupStatus,
+			'status_with_project_folder_setup_status' => $appPasswordSetStatus
```
individual-it, line 572: maybe we need other names, I got confused - see comment above

integration_setup.sh
```diff
@@ -146,26 +166,51 @@ create_storage_response=$(curl -X POST -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWOR
...
   if [[ ${error_id} == "urn:openproject-org:api:v3:errors:MultipleErrors" ]]; then
-    error_messages=$(echo $create_storage_response | jq -r "._embedded.errors[].message")
-    log_error "Got multiple errors when setting up
-      OP storage: ${OPENPROJECT_STORAGE_NAME}
-      for Nextcloud: '${NEXTCLOUD_HOST}
-      using endpoint: ${OPENPROJECT_BASEURL_FOR_STORAGE}"
-    log_error "Error Message(s): ${error_messages}"
-    log_info "You could try deleting the file storage in OpenProject and run the script again."
+    # if files storage is already created with the provided host and storage name
+    # we assume that the integration set up is done in both application
```
individual-it, line 176: I don't see where we check the correct name and host, or is OpenProject doing that?

integration_setup.sh
```diff
@@ -146,26 +166,51 @@ create_storage_response=$(curl -X POST -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWOR
...
+		exit 1
+    fi
+    log_success "File storage with name '${OPENPROJECT_STORAGE_NAME}' with host '${NEXTCLOUD_HOST}' has already been created on 'OpenProject'."
+    status=$(isNextcloudAdminConfigOk)
+    if [[ "$status" -ne 0 ]]; then
+    	log_error "Some admin config for integration is missing on Nextcloud '${NEXTCLOUD_HOST}'."
+    	if [[ ${SETUP_PROJECT_FOLDER} == 'true' ]]; then
+    		log_error "Or project folder setup might be missing on Nextcloud '${NEXTCLOUD_HOST}'."
+    	fi
+    	log_info "You could try deleting the file storage '${OPENPROJECT_STORAGE_NAME}' in OpenProject and run the script again."
```
individual-it, line 202: This will happen when the setup in Nextcloud is not complete or wrong, how will deleting the file-storage in OpenProject help?

lib/Service/OauthService.php
SagarGi: https://github.com/nextcloud/integration_openproject/pull/638 A separate PR has been made why this change was necessary!

integration_setup.sh
SagarGi: I have renamed it. Please check if it makes sense now!

integration_setup.sh
SagarGi: Yes in case of any error we need a manual intervention so we need to do it from start and the openproject storage needs to be deleted.

integration_setup.sh
SagarGi: I have checked the messaged instead of the attribute from the response which is more readable.

integration_setup.sh
```diff
@@ -146,26 +166,52 @@ create_storage_response=$(curl -X POST -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWOR
...
+    readarray -t error_messages <<<"$error_messages_grep"
+    count=0
+    for element in "${error_messages[@]}"; do
+      if [[ "$element" == "Host has already been taken." ]]; then
+          (( count +=1 ))
+      elif [[ "$element" == "Name has already been taken." ]]; then
+          (( count +=1 ))
+      fi
+    done
+    if [[ "$count" -ne 2 ]]; then
```
individual-it, line 187: What about making sure that there are only 2 error messages and make the variable names more explicit?

```suggestion
    error_count = 0
    host_already_taken = false
    name_already_taken = false
    for element in "${error_messages[@]}"; do
      if [[ "$element" == "Host has already been taken." ]]; then
          host_already_taken = true
      elif [[ "$element" == "Name has already been taken." ]]; then
          name_already_taken = true
      fi
      (( error_count +=1 ))
    done
    if [[ $host_already_taken != true || $name_already_taken != true || "$error_count" -ne 2  ]]; then
```

integration_setup.sh
```diff
@@ -146,26 +166,52 @@ create_storage_response=$(curl -X POST -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWOR
...
   if [[ ${error_id} == "urn:openproject-org:api:v3:errors:MultipleErrors" ]]; then
-    error_messages=$(echo $create_storage_response | jq -r "._embedded.errors[].message")
-    log_error "Got multiple errors when setting up
-      OP storage: ${OPENPROJECT_STORAGE_NAME}
-      for Nextcloud: '${NEXTCLOUD_HOST}
-      using endpoint: ${OPENPROJECT_BASEURL_FOR_STORAGE}"
-    log_error "Error Message(s): ${error_messages}"
-    log_info "You could try deleting the file storage in OpenProject and run the script again."
+    # if files storage is already created with the provided host and storage name
+    # we assume that the integration set up is done in both application
```
individual-it, line 176:
```suggestion
    # if the files storage is already created with the provided host and storage name
    # we assume that the integration set up is done in both application
    # to check that, we parse the error messages
    # if there are only two error messages and those are about the host and name being taken
    # we assume the setup is complete with the given parameters
```

integration_setup.sh
```diff
@@ -285,7 +329,7 @@ EOF
               deleteOPStorageAndPrintErrorResponse "$save_app_password_response"
               exit 1
         fi
-    	log_info "success!"
+    	  log_success "Saving 'OpenProject' user application password to OpenProject was successful."
```
individual-it, line 332: wrong indent?

lib/Controller/ConfigController.php
```diff
@@ -557,6 +557,22 @@ public function checkConfig(): DataResponse {
...
+	/**
+	 * @NoCSRFRequired
+	 * @return DataResponse
+	 */
+	public function checkAdminConfigOk(): DataResponse {
+		$adminConfigStatusWithoutGroupFolderSetupStatus = OpenProjectAPIService::isAdminConfigOk($this->config);
+		$appPasswordSetStatus = $this->openprojectAPIService->hasAppPassword();
+		// admin config completion can be checked in 2 phase
+		// 1. with project folder setup (with app password)
+		// 2. without project folder setup (no app password)
```
individual-it, line 569: isn't that the other way around that first the app if configured without project folder and then the project folder is added?
Or is it not about the order? Then "phase" is misleading. You could say: "there are two parts of the app configuration", one needs always to be setup, the other is optional

individual-it (changes requested):

lib/Controller/ConfigController.php
SagarGi: No its not about the order. Yes i can change the comments.

individual-it (approved):

PR #643, opened 2024-06-03
[WP#55363]Update script according to extended `OpenProject` API

SagarGi:
## Description
This PR:
- Adds check for the file storage completion in `OpenProject` when we are rerunning the script again with the same storage name.
- Adds check for checking OpenProject version >= 14.2.0 in the script, from now the script will only run for OP version >= 14.2.0  

- Fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/55363

## Types of changes

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/integration_setup.sh
+++ b/integration_setup.sh
@@ -82,6 +82,8 @@ openproject_host_state_response=$(curl -s -X GET -u${OP_ADMIN_USERNAME}:${OP_ADM
 # these two data are set to "false" when the integration is done without project folder setup
 setup_project_folder=false
 setup_app_password=false
+# Minimum OpenProject version required in order to run this script.
+MINIMUM_OP_VERSION="14.2.0"
 
 isNextcloudAdminConfigOk() {
 	admin_config_response=$(curl -s -X GET -u${NC_ADMIN_USERNAME}:${NC_ADMIN_PASSWORD} \
@@ -104,6 +106,28 @@ isNextcloudAdminConfigOk() {
 	fi
 }
 
+isOpenProjectFileStorageConfigOk() {
+	all_file_storages_available_response=$(curl -s -X GET -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} \
+      ${OPENPROJECT_BASEURL_FOR_STORAGE} \
+      -H 'accept: application/hal+json' \
+      -H 'Content-Type: application/json' \
+      -H 'X-Requested-With: XMLHttpRequest'
+    )
+    oauth_configured_status=$(echo $all_file_storages_available_response | jq -r --arg name "$OPENPROJECT_STORAGE_NAME" '.["_embedded"].elements[] | select(.name == $name) | .configured')
+    has_nc_application_password_status=$(echo $all_file_storages_available_response | jq -r --arg name "$OPENPROJECT_STORAGE_NAME" '.["_embedded"].elements[] | select(.name == $name) | .hasApplicationPassword')
+	if [[ ${SETUP_PROJECT_FOLDER} == 'true' ]]; then
+		if [[ ${oauth_configured_status} == 'true' && ${has_nc_application_password_status} == 'true' ]]; then
+			echo 0
+		else
+			echo 1
+		fi
+	elif [[ ${oauth_configured_status} == 'true' ]]; then
+		echo 0
+	else
+		echo 1
+	fi
+}
+
 # check if both instances are started or not
 if [[ $(echo $openproject_host_state_response | jq -r "._type") != "Configuration" ]]; then
   if [[ $(echo $openproject_host_state_response | jq -r ".errorIdentifier") == "urn:openproject-org:api:v3:errors:Unauthenticated" ]]; then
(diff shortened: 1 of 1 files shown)


individual-it (commented):
what happens if the script is run with an older version of OP, that does not have the new API endpoint?
Do we need to be able to run the script with older versions of OP?

SagarGi:
> what happens if the script is run with an older version of OP, that does not have the new API endpoint? Do we need to be able to run the script with older versions of OP?

@individual-it  For this script we need `OpenProject` version `14.1.1` specifically.I totally forgot about that. May be we need to check the version in the script itself otherwise it will give error.

individual-it:
or we run the script on older versions without the new check

SagarGi:
> or we run the script on older versions without the new check

yes can be done. But anyways needs the `OpenProject` Version to be fetched!

integration_setup.sh
```diff
@@ -199,17 +233,31 @@ if [[ ${response_type} == "Error" ]]; then
...
+		status_op=$(isOpenProjectFileStorageConfigOk)
+		if [[ "$status_op" -ne 0 ]]; then
+			log_error "File storage '$OPENPROJECT_STORAGE_NAME' config is incomplete on OpenProject '${OPENPROJECT_HOST}' for integration with Nextcloud."
+			if [[ ${SETUP_PROJECT_FOLDER} == 'true' ]]; then
+				log_error "Or the application password has not been set on 'OpenProject' '${OPENPROJECT_HOST}'."
+			fi
+			log_info "You could try deleting the file storage '${OPENPROJECT_STORAGE_NAME}' in OpenProject and run the script again."
+			exit 1
+		fi
+		log_success "File storage name '$OPENPROJECT_STORAGE_NAME' on OpenProject for integration has already been set up."
```
individual-it, line 249: is it `in OpenProject` or `on OpenProject`? maybe @phil-davis can  give a hint

integration_setup.sh
```diff
@@ -199,17 +233,31 @@ if [[ ${response_type} == "Error" ]]; then
 		log_info "You could try deleting the file storage in OpenProject and run the script again."
 		exit 1
     fi
-    log_success "File storage with name '${OPENPROJECT_STORAGE_NAME}' with host '${NEXTCLOUD_HOST}' has already been created on 'OpenProject'."
-    status=$(isNextcloudAdminConfigOk)
-    if [[ "$status" -ne 0 ]]; then
-    	log_error "Some admin config for integration is missing on Nextcloud '${NEXTCLOUD_HOST}'."
+    log_success "File storage name '$OPENPROJECT_STORAGE_NAME' on OpenProject exist already."
+    openproject_version=$(getOpenProjectVersion)
+    if [[ ${openproject_version} > "14.1.2" ]]; then
```
individual-it, line 238: lets check this at the very beginning of the script and check for >= 14.2

individual-it (changes requested):

integration_setup.sh
phil-davis: "in OpenProject" sounds better - that is already done at line 246.
Other text, like at lines 242, 244 and 249, can be changed.

SagarGi:
hi @individual-it  I have create a WP for improving the log messages https://community.openproject.org/projects/nextcloud-integration/work_packages/55514 . That can be changed in the separate PR.

SagarGi:
@nabim777 can you check this script manually in older version and also `14.2.0`  of OpenProject?

integration_setup.sh
```diff
@@ -120,6 +142,14 @@ if [[ ${NEXTCLOUD_HOST_INSTALLED_STATE} != "true" || ${NEXTCLOUD_HOST_MAINTENANC
   exit 1
 fi
 
+# This script requires OpenProject Version >= 14.2.0
+openproject_info_response=$(curl -s -X GET -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} ${OPENPROJECT_HOST}/api/v3)
+openproject_version=$(echo $openproject_info_response | jq -r ".coreVersion")
+if [[ "$openproject_version" < "14.2.0" ]]; then
```
individual-it, line 148: also this should not be hidden in the depths of the script, but somewhere on the top where is will be easy to find and adjust

SagarGi: sure i will make variable `MINIMUM_OP_VERSION` or name similar to it in ooder to make it work.

integration_setup.sh
```diff
@@ -120,6 +144,14 @@ if [[ ${NEXTCLOUD_HOST_INSTALLED_STATE} != "true" || ${NEXTCLOUD_HOST_MAINTENANC
   exit 1
 fi
 
+# This script requires OpenProject Version >= 14.2.0
```
phil-davis, line 147:
```suggestion
# This script requires OpenProject Version >= MINIMUM_OP_VERSION
```

If we mention 14.2.0 in a comment, then it will be easy to forget to update it in the future.

integration_setup.sh
```diff
@@ -120,6 +144,14 @@ if [[ ${NEXTCLOUD_HOST_INSTALLED_STATE} != "true" || ${NEXTCLOUD_HOST_MAINTENANC
...
+openproject_info_response=$(curl -s -X GET -u${OP_ADMIN_USERNAME}:${OP_ADMIN_PASSWORD} ${OPENPROJECT_HOST}/api/v3)
+openproject_version=$(echo $openproject_info_response | jq -r ".coreVersion")
+if [[ "$openproject_version" < "$MINIMUM_OP_VERSION" ]]; then
+	log_error "This script requires OpenProject Version greater or equals to 14.2.0 but found version '$openproject_version'"
```
phil-davis, line 151:
```suggestion
	log_error "This script requires OpenProject Version greater than or equal to 14.2.0 but found version '$openproject_version'"
```

phil-davis (commented):
IMO there is another PR coming after this that has various log_ messages reorganised and rewritten, so I didn't review the detail of those messages here.

integration_setup.sh
phil-davis: And the constant `MINIMUM_OP_VERSION` needs to be substituted instead of hardcoded `14.2.0`

phil-davis (approved):

individual-it (approved):

nabim777:
> @nabim777 can you check this script manually in older version and also `14.2.0` of OpenProject?

Checked in release openproject 14.1/14.2 :heavy_check_mark:

nabim777 (approved):
LGTM :+1:

PR #694, opened 2024-09-05
[WP#57654]switch encryption for oauth secret

SagarGi:
## Description
In OAuth, now the client secret are hashed for greater security which is done in this PR https://github.com/nextcloud/server/pull/47635. Previously we just used to encrypt it.

This PR:
- Makes changes for the hash update for different NC versions for our APP
- Also remove `client_secret` when fetching the client info for the existing user.  

- Fixes https://community.openproject.org/projects/nextcloud-integration/work_packages/57654

## Types of changes

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/Service/OauthService.php
+++ b/lib/Service/OauthService.php
@@ -58,18 +58,20 @@ class OauthService {
 		$client->setName($name);
 		$client->setRedirectUri(sprintf($redirectUri, $clientId));
 		$secret = $this->secureRandom->generate(64, self::validChars);
-		if (version_compare(OC_Util::getVersionString(), '30.0.0') >= 0) {
-			$encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
-		} elseif (version_compare(OC_Util::getVersionString(), '29.0.7') >= 0 && version_compare(OC_Util::getVersionString(), '30.0.0') < 0) {
-			$encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
-		} elseif (version_compare(OC_Util::getVersionString(), '28.0.10') >= 0 && version_compare(OC_Util::getVersionString(), '29.0.0') < 0) {
-			$encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
-		} elseif (version_compare(OC_Util::getVersionString(), '27.1.11.8') >= 0 && version_compare(OC_Util::getVersionString(), '28.0.0') < 0) {
-			$encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
-		} elseif (version_compare(OC_Util::getVersionString(), '27.0.0') === 0) {
-			$encryptedSecret = $secret;
-		} else {
-			$encryptedSecret = $this->crypto->encrypt($secret);
+		$nextcloudVersion = OC_Util::getVersionString();
+		switch (true) {
+			case version_compare($nextcloudVersion, '30.0.0') >= 0:
+			case version_compare($nextcloudVersion, '29.0.7') >= 0 && version_compare($nextcloudVersion, '30.0.0') < 0:
+			case version_compare($nextcloudVersion, '28.0.10') >= 0 && version_compare($nextcloudVersion, '29.0.0') < 0:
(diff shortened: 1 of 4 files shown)


lib/Service/OauthService.php
```diff
@@ -58,10 +58,18 @@ public function createNcOauthClient(string $name, string $redirectUri): array {
...
+			$encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
+		} elseif (version_compare(OC_Util::getVersionString(), '28.0.10') >= 0 && version_compare(OC_Util::getVersionString(), '29.0.0') < 0) {
+			$encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
+		} elseif (version_compare(OC_Util::getVersionString(), '27.1.11.8') >= 0 && version_compare(OC_Util::getVersionString(), '28.0.0') < 0) {
+			$encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
+		} elseif (version_compare(OC_Util::getVersionString(), '27.0.0') === 0) {
 			$encryptedSecret = $secret;
+		} else {
+			$encryptedSecret = $this->crypto->encrypt($secret);
 		}
```
individual-it, line 73: would a switch / case statement look cleaner?
```suggestion
		To convert the `if` statement to a `switch` case statement, we need to use a combination of `true` as the switch expression and `case` statements with conditions. Here is the simplified version using a `switch` statement:

```php
switch (true) {
    case version_compare(OC_Util::getVersionString(), '30.0.0') >= 0:
    case version_compare(OC_Util::getVersionString(), '29.0.7') >= 0 && version_compare(OC_Util::getVersionString(), '30.0.0') < 0:
    case version_compare(OC_Util::getVersionString(), '28.0.10') >= 0 && version_compare(OC_Util::getVersionString(), '29.0.0') < 0:
    case version_compare(OC_Util::getVersionString(), '27.1.11.8') >= 0 && version_compare(OC_Util::getVersionString(), '28.0.0') < 0:
        $encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
        break;
    case version_compare(OC_Util::getVersionString(), '27.0.0') === 0:
        $encryptedSecret = $secret;
        break;
    default:
        $encryptedSecret = $this->crypto->encrypt($secret);
        break;
}
```

individual-it (commented):
there is more logic in createNcOauthClient so need to create unit tests for it

lib/Service/OauthService.php
```diff
@@ -48,17 +48,11 @@ public function __construct(ClientMapper $clientMapper,
...
+	 * @return string
 	 */
-	public function createNcOauthClient(string $name, string $redirectUri): array {
-		$clientId = $this->secureRandom->generate(64, self::validChars);
-		$client = new Client();
-		$client->setName($name);
-		$client->setRedirectUri(sprintf($redirectUri, $clientId));
-		$secret = $this->secureRandom->generate(64, self::validChars);
-		$nextcloudVersion = OC_Util::getVersionString();
+	public function getHashedOrEncryptedSecretBasedOnNextcloudVersions(string $secret, string $nextcloudVersion): string {
```
individual-it, line 55:
```suggestion
	public function hashOrEncryptSecretBasedOnNextcloudVersion(string $secret, string $nextcloudVersion): string {
```

`get` sounds like we are reading it from somewhere, what is not true

tests/lib/Service/OauthSeviceTest.php
```diff
@@ -0,0 +1,123 @@
+<?php
+/**
+ * @copyright Copyright (c) 2022 Swikriti Tripathi <swikriti@jankaritech.com>
+ *
+ * @author Your name <swikriti@jankaritech.com>
```
individual-it, line 5: I don't think @SwikritiT was involved in this ;-)

individual-it (changes requested):

individual-it (approved):

SagarGi:
API test is failing because of the flaky test https://github.com/nextcloud/integration_openproject/actions/runs/10729539649.

PR #706, opened 2024-10-01
tests for ci

nabim777:
for CI tests only


SagarGi:
This was just for CI check if it was failing in NC-master31. So closing it.

PR #707, opened 2024-10-03
Update script according to storage API response of openproject

SagarGi:
## Description
For creating file storage in `OpenProject`, it can either give a `MultipleErrors` or a  single `PropertyConstraintViolation` for the same error. Previously our script only used to check `MultipleErrors` but this PR updates the script to check and handle the `PropertyConstraintViolation` error as well.

>***NOTE:*** The main check is for `Nextcloud host name` and the `OpenProject file storage name` since we only provide this two property on the API request for the integration setup configuration

## Related Issue or Workpackage
https://community.openproject.org/projects/nextcloud-integration/work_packages/57674/activity

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Updated `CHANGELOG.md` file

PR #709, opened 2024-10-07
Update script according to storage API response of openproject (#707)

SagarGi:
### Description
This PR forward port the PR : https://github.com/nextcloud/integration_openproject/pull/707 and https://github.com/nextcloud/integration_openproject/pull/711 to `master` branch

### DIFF
--- a/integration_setup.sh
+++ b/integration_setup.sh
@@ -134,8 +134,12 @@ logUnhandledError() {
 	log_info "You could try deleting the file storage '${OPENPROJECT_STORAGE_NAME}' in OpenProject and run the script again."
 }
 
-checkForOpenProjectOrNextcloudIntegrationConfiguration() {
-	# This function check whether the configuration is complete or not in both OpenProject and Nextcloud
+logCompleteIntegrationConfiguration() {
+	log_success "Setup of OpenProject and Nextcloud is complete."
+	exit 0
+}
+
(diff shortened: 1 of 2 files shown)


integration_setup.sh
```diff
@@ -128,6 +128,40 @@ isOpenProjectFileStorageConfigOk() {
   fi
 }
 
+logUnhandledError() {
+	log_error "Unhandled error while creating the file storage '${OPENPROJECT_STORAGE_NAME}'"
+	log_error "OpenProject returned the following error: '${error_message}'"
+	log_info "You could try deleting the file storage '${OPENPROJECT_STORAGE_NAME}' in OpenProject and run the script again."
+}
+
+checkForOpenProjectOrNextcloudIntegrationConfiguration() {
```
individual-it, line 137:
```suggestion
checkForOpenProjectAndNextcloudIntegrationConfiguration() {
```

Or even better split into two functions, one for OP, one for NC. Every function just does one thing

individual-it (changes requested):

integration_setup.sh
SagarGi: Ah sure! I will may be first do that on release/2.7 and then do it here again.

nabim777 (approved):
LGTM if CI :green_heart:

individual-it (approved):

PR #719, opened 2024-10-16
Safer getNextcloudVersion() using the version array

julien-nc:
Checking the version against the "versionString" can fail (in Nextcloud Enterprise for example, it is "29.0.7 Enterprise"). So it makes the `version_compare` give unexpected results.
It is safer to use the version array so we always get the same clean format for the version.

cc @wielinde

### DIFF
--- a/lib/ServerVersionHelper.php
+++ b/lib/ServerVersionHelper.php
@@ -13,12 +13,14 @@ class ServerVersionHelper {
 	 * @return string
 	 */
 	public static function getNextcloudVersion(): string {
-		// for nextcloud above 31 OC_Util::getVersionString() method does not exists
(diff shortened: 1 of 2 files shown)


SagarGi (approved):
LGTM :+1:

SagarGi:
We do have a version get in here as well. I will push a commit to change it here as well https://github.com/nextcloud/integration_openproject/blob/9da5f97562f4a749896ec2d18ad2e7dd46641856/lib/Service/OauthService.php#L83

SagarGi:
In that case i think we should fix it in `release/2.7` branch and have a patch release for it. since our app has different version check condition.
CC @individual-it @julien-nc

SagarGi (approved):

PR #723, opened 2024-10-25
Fix version comparison

nabim777:
## Description

The method getVersion() returns with 4 digit array Example: [27,0,0,7] which fails on the line https://github.com/nextcloud/integration_openproject/blob/master/lib/Service/OauthService.php#L62
This error is seen on nextcloud 27 version (i.e when variable $nextcloudversion is 27.0.0)
So, this PR is for fixing such error

## Step to reproduce
1. Clone the server of branch 27

```zsh
git clone git@github.com:nextcloud/server.git -b stable27
```

2. Go to path `\server` and Checkout to the `v27.0.0`

```zsh
git checkout v27.0.0
```

3. Set up nextcloud, enable integration_opneproject apps and openproject
4. When we try to do oauth connection then the app crashes

- https://community.openproject.org/projects/nextcloud-integration/work_packages/58464

## Screen Record :
While oauth configuration on stable27 branch we got following error:
[Screencast from 10-25-2024 02:29:03 PM.webm](https://github.com/user-attachments/assets/96ca0b05-815b-4d9f-9347-69501bd2d961)

## Types of changes

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/Service/OauthService.php
+++ b/lib/Service/OauthService.php
@@ -53,13 +53,13 @@ class OauthService {
 	 */
 	public function hashOrEncryptSecretBasedOnNextcloudVersion(string $secret, string $nextcloudVersion): string {
 		switch (true) {
-			case version_compare($nextcloudVersion, '30.0.0') >= 0:
-			case version_compare($nextcloudVersion, '29.0.7') >= 0 && version_compare($nextcloudVersion, '30.0.0') < 0:
-			case version_compare($nextcloudVersion, '28.0.10') >= 0 && version_compare($nextcloudVersion, '29.0.0') < 0:
-			case version_compare($nextcloudVersion, '27.1.11.8') >= 0 && version_compare($nextcloudVersion, '28.0.0') < 0:
+			case version_compare($nextcloudVersion, '30.0.0.0') >= 0:
+			case version_compare($nextcloudVersion, '29.0.7.0') >= 0 && version_compare($nextcloudVersion, '30.0.0.0') < 0:
+			case version_compare($nextcloudVersion, '28.0.10.0') >= 0 && version_compare($nextcloudVersion, '29.0.0.0') < 0:
+			case version_compare($nextcloudVersion, '27.1.11.8') >= 0 && version_compare($nextcloudVersion, '28.0.0.0') < 0:
 				$encryptedSecret = bin2hex($this->crypto->calculateHMAC($secret));
 				break;
-			case version_compare($nextcloudVersion, '27.0.0') === 0:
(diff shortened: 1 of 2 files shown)


lib/Service/OauthService.php
```diff
@@ -53,13 +53,13 @@ public function __construct(ClientMapper $clientMapper,
...
 	public function hashOrEncryptSecretBasedOnNextcloudVersion(string $secret, string $nextcloudVersion): string {
 		switch (true) {
-			case version_compare($nextcloudVersion, '30.0.0') >= 0:
-			case version_compare($nextcloudVersion, '29.0.7') >= 0 && version_compare($nextcloudVersion, '30.0.0') < 0:
-			case version_compare($nextcloudVersion, '28.0.10') >= 0 && version_compare($nextcloudVersion, '29.0.0') < 0:
-			case version_compare($nextcloudVersion, '27.1.11.8') >= 0 && version_compare($nextcloudVersion, '28.0.0') < 0:
+			case version_compare($nextcloudVersion, '30.0.0.0') >= 0:
+			case version_compare($nextcloudVersion, '29.0.7.0') >= 0 && version_compare($nextcloudVersion, '30.0.0.0') < 0:
+			case version_compare($nextcloudVersion, '28.0.10.0') >= 0 && version_compare($nextcloudVersion, '29.0.0.0') < 0:
+			case version_compare($nextcloudVersion, '27.1.11.8') >= 0 && version_compare($nextcloudVersion, '28.0.0.0') < 0:
```
SagarGi, line 59: @julien-nc has provided some specific nextcloud version with minipatch version as well like `27.1.11.8 in work package https://community.openproject.org/projects/nextcloud-integration/work_packages/57654/activity. I think we need to be sure that the hash encryption is included for NC versions that is listed on work packages. For example in the work package its says hash is included in 30. does it mean 30.0.0.0 or 30.0.0.2 or 30.0.0.3? It is not very specific. I think may be we need to be very specific at this case? 
@individual-it

individual-it: good point, please find out the micro versions

julien-nc: * Secrets are hashed in all v30.x . The v30 release was actually 30.0.0.14
* Secrets are hashed in v29.0.7 and higher. The v29.0.7 release was 29.0.7.1
* Secrets are hashed in v28.0.10 and higher. The v28.0.10 release was 28.0.10.1
* :warning: Secrets are hashed in v27.1.11.9 and higher. The v27.1.11 release was actually 27.1.11.3 and still encrypts (does not hash yet). The enterprise release 27.1.11.9 hashes the secrets. Micro version is indeed important for 27.

SagarGi (approved):
LGTM :+1:

PR #730, opened 2024-11-06
forwardport and conflict resolved

nabim777:
forward port of branch `release2.7`


nabim777:
some unchanged appears double time so making another PR starting

PR #757, opened 2025-01-22
chore: add linting make targets

saw-jan:
## Description
Added make targets for linting php and js code, and combined targets.
Also, added use of `SERVER_PATH` env to provide server root path. This is helpful when using integration_openproject repo outside of server apps directory.

## Types of changes

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.github/workflows/shared_workflow.yml
+++ b/.github/workflows/shared_workflow.yml
@@ -92,7 +92,7 @@ jobs:
           cd server && git submodule update --init
           ./occ maintenance:install --admin-pass=admin
 
-      - name: PHP code analysis
+      - name: PHP code analysis and linting
         run: |
           # The following if block can be removed once Nextcloud no longer supports PHP 8.0
           if [ "${{matrix.phpVersion}}" -eq 8 ]; then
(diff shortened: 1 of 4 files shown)


.github/workflows/shared_workflow.yml
```diff
@@ -92,26 +92,21 @@ jobs:
...
           # The following if block can be removed once Nextcloud no longer supports PHP 8.0
           if [ "${{matrix.phpVersion}}" -eq 8 ]; then
             # Ignoring the pact test for php 8.0
             sed -i '/<ignoreFiles>/a \ <file name="tests/lib/Service/OpenProjectAPIServiceTest.php"/>' psalm.xml
           fi
           make psalm
-
-      - name: PHP code style
-        run: composer run cs:check || ( echo 'Please run `composer run cs:fix` to format your code' && exit 1 )
+          make csfixer || ( echo 'Please run `make csfixer-fix` to format your code' && exit 1 )
```
nabim777, line 103:
```suggestion
          make csfixer || ( echo 'Please run `make csfixer-fix` to format your code' && exit 1 )
```

Why keep linting in the same step? For simpler debugging and maintenance, wouldn't it be better to move it to a separate step as before?

makefile
```diff
@@ -63,27 +63,57 @@ npm-dev:
...
+	composer run cs:check
+
+.PHONY: phpcs-fix
+phpcs-fix:
+	composer run cs:fix
+
+.PHONY: lint-php
+lint-php: psalm phpcs
+
+.PHONY: lint-php
```
nabim777, line 79:
```suggestion
.PHONY: lint-php-fix
```

.github/workflows/shared_workflow.yml
saw-jan: The readability and functionality would not be compromised. we really don't care the output if the step passes but when one of the command fails then it should be pretty similar to the previous one.
if `make psalm` fails then it will show the necessary error and the next command won't run.

nabim777 (approved):
LGTM :+1:

PR #763, opened 2025-01-30
test: refactor OpenProjectAPIService class unit tests

saw-jan:
## Description
- create a construct params helper to be able to modify any param easily
- renamed `getServiceMock` -> `getOpenProjectAPIServiceMock`

## Types of changes

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/lib/Service/OpenProjectAPIServiceTest.php
+++ b/tests/lib/Service/OpenProjectAPIServiceTest.php
@@ -630,7 +630,13 @@ class OpenProjectAPIServiceTest extends TestCase {
 	}
 
 	/**
-	 * @param array<string, object> $constructParams
+	 * generates a list mocks that can be given as arguments to the constructor of OpenProjectAPIService
+	 * by default only empty mocks are generated, but specific mocks can be passed in using the
+	 * $constructParams parameter.
+	 *
+	 * Format has to be [<string> => <object>] with the first being the constructor parameter name and the second one the mock.
+	 * Example: ['avatarManager' => $createMockObject]
+	 * @param array<string, object> $constructParams specific mocks for the constructor of OpenProjectAPIService
 	 *
 	 * @return array
 	 */
@@ -658,8 +664,12 @@ class OpenProjectAPIServiceTest extends TestCase {
 			'exchangedTokenRequestedEventHelper' => $this->createMock(ExchangedTokenRequestedEventHelper::class),
 		];
 
-		// replace the mock
+		// replace default mocks with manually passed in mocks
 		foreach ($constructParams as $key => $value) {
+			if (!array_key_exists($key, $constructArgs)) {
+				throw new \InvalidArgumentException("Invalid construct parameter: $key");
+			}
+
 			$constructArgs[$key] = $value;
 		}
 
--- a/tests/lib/Service/OpenProjectAPIServiceTest.php
+++ b/tests/lib/Service/OpenProjectAPIServiceTest.php
@@ -630,6 +630,43 @@ class OpenProjectAPIServiceTest extends TestCase {
 	}
 
 	/**
+	 * @param array<string, object> $constructParams
+	 *
+	 * @return array
+	 */
+	private function getOpenProjectAPIServiceConstructArgs(array $constructParams = []): array {
+		$constructArgs = [
+			// order should be the same as in the constructor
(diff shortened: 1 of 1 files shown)


tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -629,6 +629,43 @@ private function getStorageMock($nodeClassName = null) {
...
+			'subAdmin' => $this->createMock(ISubAdmin::class),
+			'dbConnection' => $this->createMock(IDBConnection::class),
+			'logFactory' => $this->createMock(ILogFactory::class),
+			'manager' => $this->createMock(IManager::class),
+			'exchangedTokenRequestedEventHelper' => $this->createMock(ExchangedTokenRequestedEventHelper::class),
+		];
+
+		// replace the mock
+		foreach ($constructParams as $key => $value) {
+			$constructArgs[$key] = $value;
```
individual-it, line 663: if an parameter is passed in that is not in the pre-defined `constructArgs` list (e.g. because of a typo) do you intend to just add it to the list, or do we need a check here?

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -638,7 +675,10 @@ private function getStorageMock($nodeClassName = null) {
 	 */
 	private function getOpenProjectAPIService(
 		$authMethod = null,
-		$storageMock = null, $oAuth2OrOidcToken = '1234567890', $baseUrl = 'https://nc.my-server.org', $userId = 'testUser'
```
individual-it, line 641: how did the code style not catch that :see_no_evil:

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -629,6 +629,43 @@ private function getStorageMock($nodeClassName = null) {
...
+			'secureRandom' => $this->createMock(ISecureRandom::class),
+			'eventDispatcher' => $this->createMock(IEventDispatcher::class),
+			'subAdmin' => $this->createMock(ISubAdmin::class),
+			'dbConnection' => $this->createMock(IDBConnection::class),
+			'logFactory' => $this->createMock(ILogFactory::class),
+			'manager' => $this->createMock(IManager::class),
+			'exchangedTokenRequestedEventHelper' => $this->createMock(ExchangedTokenRequestedEventHelper::class),
+		];
+
+		// replace the mock
```
individual-it, line 661:
```suggestion
		// replace default mocks with manually passed in mocks
```

tests/lib/Service/OpenProjectAPIServiceTest.php
```diff
@@ -629,6 +629,43 @@ private function getStorageMock($nodeClassName = null) {
 		return $storageMock;
 	}
 
+	/**
+	 * @param array<string, object> $constructParams
```
individual-it, line 633: maybe over explaining, bu this is how I understood what the function does

```suggestion
     * generates a list mocks that can be given as arguments to the constructor of OpenProjectAPIService
     * by default only empty mocks are generated, but specific mocks can be passed in using the 
     * $constructParams parameter
     * 
	 * @param array<string, object> $constructParams specific mocks for the constructor of OpenProjectAPIService. Format has to be [[<string>, <string>]] with the first being the constructor parameter name and the second one the mock
```

individual-it (changes requested):

tests/lib/Service/OpenProjectAPIServiceTest.php
saw-jan: maybe the linter is set to less strict rules.

tests/lib/Service/OpenProjectAPIServiceTest.php
saw-jan: ahh, right. need to catch that. :+1:

saw-jan: I have added the check for invalid param.

tests/lib/Service/OpenProjectAPIServiceTest.php
saw-jan: I have added the docs but they are a little bit different. Please have a look.

individual-it (approved):

PR #765, opened 2025-02-04
test: extend unit tests for Dashboard component

saw-jan:
## Description
Extended unit tests for Dashboard component

- [WP#60395](https://community.openproject.org/wp/60395)

## Types of changes

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/jest/views/__snapshots__/Dashboard.spec.js.snap
+++ b/tests/jest/views/__snapshots__/Dashboard.spec.js.snap
@@ -1,6 +1,47 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Dashboard.vue should show the notification items in the Dashboard 1`] = `
+exports[`Dashboard.vue auth method: OAUTH2 should show the notification items in the Dashboard 1`] = `
+[
+  {
+    "avatarUrl": "http://localhost/apps/integration_openproject/avatar?userId=8&userName=Admin de DEV user",
+    "avatarUsername": "Admin de DEV userz",
+    "id": "17",
+    "mainText": "(6) Create wireframes for new landing page",
+    "overlayIconUrl": "",
+    "subText": "Scrum project - assignee, mentioned, Date alert",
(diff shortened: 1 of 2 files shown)


tests/jest/views/__snapshots__/Dashboard.spec.js.snap
```diff
@@ -1,6 +1,47 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Dashboard.vue should show the notification items in the Dashboard 1`] = `
+exports[`Dashboard.vue auth method: OAUTH2 should show the notification items in the Dashboard 1`] = `
+[
+  {
+    "avatarUrl": "http://localhost/apps/integration_openproject/avatar?userId=8&userName=Admin de DEV user",
```
nabim777, line 6:
```suggestion
    "avatarUrl": "http://localhost/apps/integration_openproject/avatar?userId=8&userName=Admin de DEV user",
```
URL contain a space.  Is this a valid URL?

individual-it, line 6: > URLs cannot contain spaces. URL encoding normally replaces a space with a plus (+) sign or with %20.

https://www.w3schools.com/tags//ref_urlencode.asp

How comes that this works? Or are the avatars broken?
But that seems to be same in the old tests, so could be investigated and fixed in an other PR

individual-it (approved):

tests/jest/views/__snapshots__/Dashboard.spec.js.snap
saw-jan: Browsers does the proper treatment to the url. that's why we didn't notice

saw-jan: >But that seems to be same in the old tests, so could be investigated and fixed in an other PR

I will investigate it :+1:

PR #769, opened 2025-02-07
Group the API endpoints that contact OP in an OCS controller

julien-nc:
Hey there!

Here is a change suggestion to make it possible to use this integration app from somewhere else than the web UI.
All the endpoints that communicate with OP are grouped in an OCS controller.
OCS endpoints can be reached without having a web session but with any supported authentication method (basic auth, app password etc...).

OCS endpoints always return an `ocs` object which contains a `meta` key and a `data` one. More info at https://docs.nextcloud.com/server/latest/developer_manual/client_apis/OCS/ocs-api-overview.html

The frontend has been adjusted but not tested IRL. This does not affect how the app works.

This would allow us, in [the agency feature](https://docs.nextcloud.com/server/latest/admin_manual/ai/app_context_agent.html), to make OP-related tools. In short, NC users would be able to ask a conversational agent to create a work package in OP for example.

cc @wielinde

### DIFF
--- a/lib/Controller/OpenProjectAPIController.php
+++ b/lib/Controller/OpenProjectAPIController.php
@@ -9,78 +9,37 @@
 namespace OCA\OpenProject\Controller;
 
 use Exception;
-use GuzzleHttp\Exception\ClientException;
-use GuzzleHttp\Exception\ConnectException;
-use GuzzleHttp\Exception\RequestException;
-use GuzzleHttp\Exception\ServerException;
 use InvalidArgumentException;
 use OCA\OpenProject\AppInfo\Application;
 use OCA\OpenProject\Exception\OpenprojectErrorException;
 use OCA\OpenProject\Service\OpenProjectAPIService;
-use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
 use OCP\AppFramework\Http\DataDisplayResponse;
 use OCP\AppFramework\Http\DataDownloadResponse;
 use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\OCSController;
 use OCP\Files\NotFoundException;
 use OCP\Files\NotPermittedException;
-use OCP\Http\Client\LocalServerException;
 
 use OCP\IConfig;
 use OCP\IRequest;
-use OCP\IURLGenerator;
-use Psr\Log\LoggerInterface;
 
-class OpenProjectAPIController extends Controller {
+class OpenProjectAPIController extends OCSController {
 
-	/**
-	 * @var OpenProjectAPIService
-	 */
-	private $openprojectAPIService;
-	/**
-	 * @var string|null
-	 */
-	private $userId;
-	/**
-	 * @var string
-	 */
-	private $accessToken;
-
-	/**
-	 * @var string
-	 */
-	private $openprojectUrl;
-
-	/**
-	 * @var IConfig
-	 */
-	private $config;
-
-	/**
-	 * @var IURLGenerator
-	 */
-	private $urlGenerator;
-
-	/**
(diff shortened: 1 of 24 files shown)


lib/Controller/OpenProjectAPIController.php
```diff
@@ -112,17 +71,15 @@ private function validatePreRequestConditions(): array {
 
 	/**
 	 * get openproject instance URL
-	 * @NoAdminRequired
 	 *
 	 * @return DataResponse
 	 */
+	#[NoAdminRequired]
```
saw-jan, line 77: App's minimum Nextcloud support is 27 and  attributes are supported from nextcloud 27. So this is good.
>The attributes are only available in Nextcloud 27 or later.
https://docs.nextcloud.com/server/latest/developer_manual/basics/controllers.html#authentication

saw-jan (commented):
Thanks @julien-nc  :heart: 
Generally looks good :+1: . I will work on fixing the tests

saw-jan:
I have added wp for this one: https://community.openproject.org/wp/61359

saw-jan:
@julien-nc How much stability and carefulness would you expect in these ocs endpoints? because we might change the endpoints' behavior as per the need. but now, since these are also exposed to non-web clients, we need to be more careful than before.

julien-nc:
@saw-jan We need them to be a little bit stable :grin: . So we could keep the internal non-OCS endpoints for the UI and these can change in time. And maybe select a few of them (search a WP, create a WP, link a file to a WP for example) and duplicate them in the OCS controller so they can be more stable than the "internal/private" ones. Wdyt?

saw-jan:
> @saw-jan We need them to be a little bit stable 😁 . So we could keep the internal non-OCS endpoints for the UI and these can change in time. And maybe select a few of them (search a WP, create a WP, link a file to a WP for example) and duplicate them in the OCS controller so they can be more stable than the "internal/private" ones. Wdyt?

having separate endpoints could also be an option. but that means maintaining two versions of same thing.
So, IMO, these endpoints will not change that much and if it does then we will do with version releases.

I will work on the tests and then let's merge this one.

DaphneMuller:
@saw-jan @artonge hi everyone, we have been asked to deliver some implementations that build upon these changes within 1,5 week from now. Is there any chance you can give this a good review before the end of this week? We are under some time pressure here.

artonge (approved):
I did a quick smoke test on the dashboard, file action and sidebar tab, and it seems to work.

saw-jan:
> @saw-jan @artonge hi everyone, we have been asked to deliver some implementations that build upon these changes within 1,5 week from now. Is there any chance you can give this a good review before the end of this week? We are under some time pressure here.

Ahh, okay, we weren't aware of that. Sorry for the delay. we will work on this immediately.
thanks for letting us know

saw-jan:
Vue tests are fixed. I will update the branch once the php unit tests are aslo fixed

dominic-braeunlein:
Hi @DaphneMuller, that should be possible. We also aim to release it on Monday or, at the latest, Tuesday with v2.8, including Nextcloud 31 support.

lib/Service/OpenProjectAPIService.php
```diff
@@ -1445,7 +1445,7 @@ public function getOpenProjectUserAvatarUrl(array $entry): string {
 			: '';
 		$userId = preg_replace('/.*\//', "", $userIdURL);
 		return $this->urlGenerator->getAbsoluteURL(
-			'index.php/apps/integration_openproject/avatar?' .
+			'ocs/v1.php/apps/integration_openproject/api/v1/avatar?' .
```
saw-jan, line 1448: @julien-nc I hope this is the correct change. Please, have a look

saw-jan: Thanks!

saw-jan:
Merging the PR. Thanks everyone :heart:

DaphneMuller:
thanks for the quick reviews and nice collaboration!

PR #776, opened 2025-02-25
build: add NC compose setup to use specific server branch

saw-jan:
## Description

- use `ghcr.io/juliusknorr/nextcloud-dev-php83:master` docker image instead of `nextcloud/apache`
- following envs are available to change the image while running the setup
  - `NEXTCLOUD_IMAGE_PHP` - php version to use (default: `83`. see [available list](https://github.com/juliusknorr?tab=packages&repo_name=nextcloud-docker-dev))
  - `NEXTCLOUD_IMAGE_TAG` - image tag to use (default: `master`)
  - `NEXTCLOUD_AUTOINSTALL_APPS` - space separate list of apps to enable
- configure nextcloud branch using `NEXTCLOUD_BRANCH`

- [WP#61735](https://community.openproject.org/wp/61735)

## Types of changes

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


dev/compose.yaml
```diff
@@ -83,22 +85,28 @@ services:
...
-      PHP_UPLOAD_SIZE: '1024M'
-      POSTGRES_HOST: nextcloud-db
-      POSTGRES_DB: nextcloud
-      POSTGRES_USER: nextcloud
-      POSTGRES_PASSWORD: nextcloud
+      SQL: pgsql
+      VIRTUAL_HOST: nextcloud
+      SERVER_BRANCH: ${NEXTCLOUD_BRANCH:-stable31}
+      NEXTCLOUD_AUTOINSTALL: YES
+      NEXTCLOUD_TRUSTED_DOMAINS: nextcloud.local
```
nabim777, line 97:
```suggestion
      SQL: pgsql
      VIRTUAL_HOST: nextcloud
      SERVER_BRANCH: ${NEXTCLOUD_BRANCH:-stable31}
      NEXTCLOUD_AUTOINSTALL: YES
      NEXTCLOUD_TRUSTED_DOMAINS: nextcloud.local
      NEXTCLOUD_AUTOINSTALL_APPS:
```

Can we add this in the env? this will automatically enable the apps 
NEXTCLOUD_AUTOINSTALL_APPS

dev/compose.yaml
```diff
@@ -83,22 +85,30 @@ services:
...
+      NEXTCLOUD_AUTOINSTALL_APPS: oidc groupfolders
```
nabim777, line 98:
```suggestion
      NEXTCLOUD_AUTOINSTALL_APPS: oidc groupfolders
```

IMO it will be better if it is also added in the `.env` as we may need to auto-install other apps, like
[Terms of services](https://github.com/nextcloud/terms_of_service)
[Activity](https://github.com/nextcloud/activity)

nabim777:
I think it would be good to have all script files in one folder named `scripts` and add a comment in each script file commenting on what it does.

saw-jan:
> I think it would be good to have all script files in one folder named `scripts` and add a comment in each script file commenting on what it does.

which scripts? if you mean the new one added here, then it's not possible to group them in a separate directory other than `hooks/`. because the image we are using expects hooks to be in `/shared/hooks` with the exact same name.
So we are mapping the volume like `./nextcloud:/shared`

dev/compose.yaml
saw-jan: done.

nabim777 (approved):
LGTM :+1:

PR #812, opened 2025-04-16
Remove myself from CODEOWNERS

juliusknorr:
While I'm happy to review code from time if specific insights are needed where my feedback is wanted, having myself as a code owner here is nothing I can cope with given my current workload and rate of github notifications.

Therefore removing me to avoid getting notified about every PR.


NobodysNightmare:
Sure we can do that.

IIRC our main motivation to introduce a codeowners file was to reduce the amount of people that can write to this repository unnoticed, because we inherit over 700 people with write access from the Nextcloud organization.

saw-jan (approved):

PR #869, opened 2025-08-15
feat: allow removing user from OpenPorject group

saw-jan:
## Description
This feature allows `OpenProject` user to remove users from the `OpenProject` group.
The implementation is as follows:
1. Listen for remove user from group API requests
2. If the group is NOT `OpenProject`, let the request continue
2. If the group is `OpenProject`,
    - check if the user to be removed is in the group `OpenProjectAll` (new group where all OP users will be dumped)
    - if user is not in the new group, add it to the group and let the Delete request continue

- Resolves [66120](https://community.openproject.org/wp/66120)

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Allow OpenProject subadmin to remove users from OpenProject group [#869](https://github.com/nextcloud/integration_openproject/pull/869)
```
individual-it, line 18:
```suggestion
- Allow OpenProject subadmin to remove users from OpenProject group, by adding that user first to and other group [#869](https://github.com/nextcloud/integration_openproject/pull/869)
```
it might sound that we changed the NC behaviour

lib/AppInfo/Application.php
```diff
@@ -150,4 +157,63 @@ public function registerNavigation(IUserSession $userSession): void {
...
+	 */
+	public function listenRemoveUserFromGroupRequest(
+		IRequest $request,
+		IGroupManager $groupManager,
+		IUserManager $userManager,
+		LoggerInterface $logger,
+	): void {
+		// url path pattern to remove user from group
+		// e.g: /cloud/users/{userid}/groups   (path returned by $request->getPathInfo())
+		$urlPattern = '/.*\/users\/([^\/]+)\/groups$/';
```
individual-it, line 183: does it work reliably if we have `index.php` or not?

lib/AppInfo/Application.php
```diff
@@ -150,4 +157,63 @@ public function registerNavigation(IUserSession $userSession): void {
...
+		if ($request->getMethod() === Request::METHOD_DELETE && \preg_match($urlPattern, $request->getPathInfo())) {
+			$fromGroup = $request->getParam('groupid');
+			$userToRemove = \preg_replace($urlPattern, '$1', $request->getPathInfo());
```
individual-it, line 186: shouldn't we use `preq_match` here too? that would return all matches

tests/lib/AppInfo/ApplicationTest.php
```diff
@@ -0,0 +1,229 @@
...
+		bool $hasUser,
+		bool $existsInGroup
+	): void {
+		$userMock = null;
+		if ($hasUser) {
+			$userMock = $this->createMock(IUser::class);
+		}
+
+		$this->request->expects($this->once())->method('getMethod')->willReturn('DELETE');
+		$this->request->expects($this->any())->method('getPathInfo')->willReturn("/cloud/users/$userName/groups");
```
individual-it, line 124: will that be an int id or a string with funny characters?

lib/AppInfo/Application.php
saw-jan: `.*` should match anything. But `getPathInfo()` method will return the stripped-out paths.

```
for: /index.php/apps/files/api/v1
returns: /apps/files/api/v1

for: /ocs/v2.php/cloud/users/jane/groups?groupid=OpenProjectAll
returns: /cloud/users/jane/groups
```

lib/AppInfo/Application.php
saw-jan: there should be only one match as per the pattern and the url path that we are interested in.
so, `preg_replace` will give us what's inside `(...)` - a capture group

tests/lib/AppInfo/ApplicationTest.php
saw-jan: do you mean to test `$userName` by providing int/number (anyway will be converted to string) and funny chars?

lib/AppInfo/Application.php
```diff
@@ -150,4 +157,63 @@ public function registerNavigation(IUserSession $userSession): void {
...
+					$errorMessage = 'Group "' . Application::OPENPROJECT_ALL_GROUP_NAME . '" not found.' .
+						' This group is required before removing users from "' . Application::OPEN_PROJECT_ENTITIES_NAME . '" group.';
+					$logger->error($errorMessage);
+					throw new OCSBadRequestException($errorMessage);
+				}
+
+				$allGroup->addUser($user);
+				$logger->debug("User '$userToRemove' added to '" . Application::OPENPROJECT_ALL_GROUP_NAME . "' group.");
+			} else {
+				$logger->debug('User already exists in "' . Application::OPENPROJECT_ALL_GROUP_NAME . '" group');
```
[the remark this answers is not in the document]
saw-jan: yes, because log is in json format and needs to escape double quote. we cannot fully eliminate this but yeah, let's think about it in the future refactoring/linting.

lib/AppInfo/Application.php
```diff
@@ -51,7 +56,8 @@
  */
 class Application extends App implements IBootstrap {
 	public const APP_ID = 'integration_openproject';
-	public const  OPEN_PROJECT_ENTITIES_NAME = 'OpenProject';
+	public const OPEN_PROJECT_ENTITIES_NAME = 'OpenProject';
+	public const OPENPROJECT_ALL_GROUP_NAME = 'OpenProjectAll';
```
saw-jan, line 60:
```suggestion
	public const OPENPROJECT_ALL_GROUP_NAME = 'OpenProjectSuspended';
```

individual-it (approved):

PR #870, opened 2025-08-18
fix: use alternative getFoldersForGroups method instead of removed getFoldersForGroup method

saw-jan:
## Description
Method `getFoldersForGroup` was recently removed in the master branch of groupfolders. See https://github.com/nextcloud/groupfolders/pull/3883
Use the existing alternative method: `getFoldersForGroups` which returns the same return type and the only difference is that how we provide the group.
```diff
- getFoldersForGroup($groupId);
+ getFoldersForGroups([$groupId]);
```
Compatible with nextcloud:
- master
- stable31
- stable30

- Fixes the failing CI https://github.com/nextcloud/integration_openproject/actions/runs/17026427055/job/48262366973

## Types of changes

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/lib/Service/OpenProjectAPIServiceTest.php
+++ b/tests/lib/Service/OpenProjectAPIServiceTest.php
@@ -2225,8 +2225,8 @@ class OpenProjectAPIServiceTest extends TestCase {
 			]];
 		}
 		$folderManagerMock
-			->method('getFoldersForGroup')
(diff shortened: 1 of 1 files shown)


saw-jan:
> so now we need to check that nobody is running the old version of the group-folder app

but there will be a problem if we use the older version of the integration_op app with the new version of the groupfolders app

PR #1074, opened 2026-06-24
fix: handle invalid boolean auth-method value

saw-jan:
## Description
Do strict comparison while using `in_array` method.
Problematic code :-1: :
```php
in_array(true, ['e1', 'e2']);  // true ❌
```
Fix :+1::
```php
in_array(true, ['e1', 'e2'], true);  // false ✔️
```

- Fixes https://community.openproject.org/wp/NEX-638

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/Service/SettingsService.php
+++ b/lib/Service/SettingsService.php
@@ -102,7 +102,7 @@ class SettingsService {
 		$settingsToSkip = [];
 
 		if ($completeSetup) {
-			if (!\in_array('authorization_method', $settingsToCheck)) {
+			if (!\in_array('authorization_method', $settingsToCheck, true)) {
 				throw new InvalidArgumentException("'authorization_method' setting is missing");
 			}
 			$authMethod = $values['authorization_method'];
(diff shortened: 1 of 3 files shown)


lib/Service/SettingsService.php
Ashim-Stha:
```suggestion
			if (!\in_array('authorization_method', $settingsToCheck, true)) {
```

should we do strict comparision here as well?

Ashim-Stha: and also everywhere in the project where there is loose comparision?

nabim777:
I think we should create a work package to apply strict mode to `in_array()` checks across the codebase where needed.

lib/Service/SettingsService.php
saw-jan: >and also everywhere in the project where there is loose comparision?

I would not do that in this PR.

saw-jan: >should we do strict comparision here as well?

AI suggestion seems misleading. but yeah this case can happen here:
```php
in_array('authorization_method', [true]) // true
```

nabim777 (approved):
LGTM 👍🏻

Ashim-Stha (approved):

PR #1131, opened 2026-07-29
test(php): add phpunit tests for OauthService

saw-jan:
## Description
Add phpunit tests for `OauthService` class

- Resolves: https://community.openproject.org/wp/NEX-106

## Types of changes

## Checklist:
- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


tests/lib/Service/OauthServiceTest.php
```diff
@@ -0,0 +1,155 @@
...
+	 * @return void
+	 */
+	public function testCreateNcOauthClient(): void {
+		$client = $this->getClient();
+		$expectedClient = [
+			'id' => 1,
+			'nextcloud_oauth_client_name' => $client->getName(),
+			'openproject_redirect_uri' => $client->getRedirectUri(),
+			'nextcloud_client_id' => $client->getClientIdentifier(),
+			'nextcloud_client_secret' => $client->getClientIdentifier(),
```
Ashim-Stha:
```suggestion
			'nextcloud_client_secret' => $client->getSecret(),
```

saw-jan: the function returns the plain secret not the encoded one. And because of ISecureRandom mock, we get same value for id and secret

tests/lib/Service/OauthServiceTest.php
```diff
@@ -0,0 +1,155 @@
...
+		$clientMapperMock->expects($this->once())->method('getByUid')
+			->with(1)
+			->willReturn($client);
+		$clientMapperMock->expects($this->once())->method('update')
+			->with($client);
+
+		$oauthService = $this->getOauthServiceMock($clientMapperMock);
+
+		$result = $oauthService->setClientRedirectUri(1, $client->getRedirectUri());
+		$this->assertEquals(true, $result);
```
nabim777:
```suggestion
		$this->assertTrue($result);
```

tests/lib/Service/OauthServiceTest.php
```diff
@@ -0,0 +1,155 @@
...
+		$clientMapperMock = $this->createMock(ClientMapper::class);
+		$clientMapperMock->expects($this->once())->method('getByUid')
+			->with(1)
+			->willThrowException(new ClientNotFoundException());
+		$clientMapperMock->expects($this->never())->method('update');
+
+		$oauthService = $this->getOauthServiceMock($clientMapperMock);
+
+		$result = $oauthService->setClientRedirectUri(1, $client->getRedirectUri());
+		$this->assertEquals(false, $result);
```
nabim777:
```suggestion
		$this->assertFalse($result);
```

tests/lib/Service/OauthServiceTest.php
```diff
@@ -0,0 +1,155 @@
...
+
+		$clientMapperMock = $this->createMock(ClientMapper::class);
+		$clientMapperMock->expects($this->once())->method('insert')
+			->with($this->isInstanceOf(Client::class))
+			->willReturn($client);
+
+		$oauthService = $this->getOauthServiceMock($clientMapperMock);
+
+		$clientInfo = $oauthService->createNcOauthClient($client->getName(), $client->getRedirectUri());
+		$this->assertEquals($expectedClient, $clientInfo);
```
nabim777:
```suggestion
		$this->assertSame($expectedClient, $clientInfo);
```

tests/lib/Service/OauthServiceTest.php
```diff
@@ -0,0 +1,155 @@
...
+	/**
+	 * @return Client
+	 */
+	protected function getClient(): Client {
+		$client = new Client();
+		$client->setId(1);
+		$client->setName('Test Client');
+		$client->setRedirectUri('https://example.com/callback');
+		$client->setClientIdentifier('randomString');
+		$client->setSecret('randomString');
```
nabim777, line 51: Why are these values the same? Are they supposed to be different?

tests/lib/Service/OauthServiceTest.php
saw-jan: need to do strict assertion :+1:

tests/lib/Service/OauthServiceTest.php
saw-jan: because we are mocking random `generate` method to return `randomString`.

saw-jan: but it can be anything here. secrets won't be used in the test

nabim777: I was a bit confused because both values are the same.

If the specific values don't matter for this test, could we use different strings instead? It would make it clearer that clientIdentifier and secret are separate values.

saw-jan: I don't think so, because this a test client for assertion and we are mocking `generate` method which will always return same.

Ashim-Stha (approved):

nabim777 (approved):
LGTM 👍🏻

